PR #17, opened 2023-04-09
feat: implement time module

RonnieSalomonsen:
Port of the time module.

I've implemented time.now(), no rule changes should be needed. Used @Blacksyke magic module as template for this module. @wxsBSD and @Blacksyke decided it was best to add a minimal protobuf specification rather than hack the build scripts to account for modules that only have exported functions and return no data, like this one.

```
cargo test --package yara-x --lib --features time-module -- modules::time::tests --nocapture

    Finished test [unoptimized + debuginfo] target(s) in 1.76s
     Running unittests src/lib.rs (target/debug/deps/yara_x-56a7c9914ffab9dd)

running 1 test
test modules::time::tests::e2e_test ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 32 filtered out; finished in 0.19s
```


yara-x/src/modules/time.rs
```diff
@@ -0,0 +1,41 @@
...
+
+#[module_main]
+fn main(_ctx: &ScanContext) -> Time {
+    // Nothing to do, but we have to return our protobuf
+    let time_proto = Time::new();
+    time_proto
+}
+
+#[module_export(name = "now")]
+fn time_now(ctx: &ScanContext) -> Option<i64> {
```
plusvic, line 13: You don't need to use `name ="now"`, naming the function `now` is enough, like this...

```rust
#[module_export]
fn now(ctx: &ScanContext) -> Option<i64> { ... }
```

RonnieSalomonsen: Thanks, will change to:

```
#[module_export]
fn now(ctx: &ScanContext) -> Option<i64> {
```

RonnieSalomonsen: Done

plusvic:
Also, don't forget to run `rustfmt` on the source code before committing. My advice is configuring your IDE to do so automatically on file save.

RonnieSalomonsen:
Thank. I have setup IDE to run rust-lang.rust-analyzer on save. Should be fixed now. Sorry about that.

PR #21, opened 2023-04-19
feat: implemented the `string` module

RonnieSalomonsen:
Hi,

Ported the String Module. Please peer review.

```
cargo test --package yara-x --lib --features string-module -- modules::string --nocapture

    Finished test [unoptimized + debuginfo] target(s) in 5.95s
     Running unittests src/lib.rs (target/debug/deps/yara_x-ff5dd03397465a33)

running 1 test
test modules::string::tests::end2end ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 32 filtered out; finished in 0.22s
```


yara-x/src/modules/string.rs
```diff
@@ -0,0 +1,86 @@
...
+    // Nothing to do, but we have to return our protobuf
+    let string_proto = String::new();
+    string_proto
+}
+
+#[module_export(name = "to_int")]
+fn to_int(ctx: &ScanContext, string: RuntimeString) -> Option<i64> {
+    let string = string.as_bstr(ctx);
+
+    match string.to_str() {
```
plusvic, line 15: Here you can take advantage of `Result::ok()` and the `?` operator, for example:

```rust
#[module_export(name = "to_int")]
fn to_int(ctx: &ScanContext, string: RuntimeString) -> Option<i64> {
    let string = string.as_bstr(ctx).to_str().ok()?;
    string.parse::<i64>().ok()
}
```

The `.ok()` method takes a `Result<T, E>` and returns an `Option<T>` if no error occurred, or `None` if otherwise. The `?` operator propagates the `None` to the caller, and returns `T`.

So, the first line returns the `&str` you need in a more compact way, while the second line parses the `i64` and returns its, or returns `None` if an error occurred.

I also think it's worth adding a `.to_str()` method to `RuntimeString`, it would be more legible:

```rust
fn to_int(ctx: &ScanContext, string: RuntimeString) -> Option<i64> {
    let string = string.to_str().ok()?;
    string.parse::<i64>().ok()
}
```

yara-x/src/modules/string.rs
```diff
@@ -0,0 +1,86 @@
...
+use crate::modules::protos::string::*;
+
+#[module_main]
+fn main(_ctx: &ScanContext) -> String {
+    // Nothing to do, but we have to return our protobuf
+    let string_proto = String::new();
+    string_proto
+}
+
+#[module_export(name = "to_int")]
```
plusvic, line 11: Don't use `name = "to_int"` if the Rust function is already named `to_int`.

yara-x/src/modules/string.rs
```diff
@@ -0,0 +1,86 @@
...
+    ctx: &ScanContext,
+    string: RuntimeString,
+    base: i64,
+) -> Option<i64> {
+    let string = string.as_bstr(ctx);
+
+    if base < 2 || base > 36 {
+        return None;
+    }
+    let strobj = match string.to_str() {
```
plusvic, line 32: Similarly, this could be simplified by using `?` and `.ok()`

```rust
fn to_int_base(
    ctx: &ScanContext,
    string: RuntimeString,
    base: i64,
) -> Option<i64> {
    if base < 2 || base > 36 {
        return None;
    }
    let string = string.as_bstr(ctx).to_str().ok()?;
    i64::from_str_radix(string, base as u32).ok()
}
```

Notice that you don't need to use `.try_into().unwrap()`, firstly because you already made sure that `base64` is between 2 and 36, so it's safe to convert it to an `u32`, and secondly because `base.try_into().unwrap()` is not safer than `base as u32`, both will panic if the conversion is not possible. `.try_into()` makes sense when you handle the error.

yara-x/src/modules/string.rs
plusvic: I've added a `to_str()` method to `RuntimeString`. This is possible now:

```
    let string = string.to_str(ctx).ok()?;
    string.parse::<i64>().ok()
```

yara-x/src/modules/string.rs
RonnieSalomonsen: Ok, i will resolve this.

yara-x/src/modules/string.rs
RonnieSalomonsen: Thanks for the feedback. I will get it fixed up.

yara-x/src/modules/string.rs
RonnieSalomonsen: Thanks. I will get it fixed.

vthib: It's not true for the `as`, this is a truncation as would be done in C. So a safe way to cast would be with a try_into.

```
let base: u32 = base.try_into().ok()?;
if base < 2 || base > 36 {
    return None;
}
```

yara-x/src/modules/string.rs
RonnieSalomonsen: Fixed

yara-x/src/modules/string.rs
RonnieSalomonsen: Fixed to @plusvic suggestion. Any objections to this @plusvic ? Else i can change it to that

yara-x/src/modules/string.rs
RonnieSalomonsen: Fixed. Thanks

yara-x/src/modules/string.rs
RonnieSalomonsen: Updated code to 
```
    let base: u32 = base.try_into().ok()?;
    if !(2..=36).contains(&base) {
```

yara-x/src/modules/string.rs
```diff
@@ -0,0 +1,72 @@
...
+fn to_int_base(
+    ctx: &ScanContext,
+    string: RuntimeString,
+    base: i64,
+) -> Option<i64> {
+    let base: u32 = base.try_into().ok()?;
+    if !(2..=36).contains(&base) {
+        return None;
+    }
+    let string = string.as_bstr(ctx).to_str().ok()?;
```
plusvic, line 26: Here you can use `string.to_str(ctx).ok()?;`

yara-x/src/modules/string.rs
```diff
@@ -0,0 +1,86 @@
...
+#[module_export]
+fn length(ctx: &ScanContext, string: RuntimeString) -> Option<i64> {
+    return Some(string.as_bstr(ctx).len().try_into().unwrap());
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn end2end() {
+        let rules = crate::compiler::Compiler::new()
```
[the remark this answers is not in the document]
RonnieSalomonsen: ok, i will fix

PR #43, opened 2023-09-29
feat: implement dylib parsing

latonis:
Implemented dylib parsing
- added tests as well
- rustfmt ran on macho module directory
- `rustfmt.toml` had an invalid value of "max", changed to "Max"


yara-x/src/modules/macho/mod.rs
```diff
@@ -839,6 +896,37 @@ fn parse_load_command(input: &[u8]) -> IResult<&[u8], LoadCommand> {
     Ok((input, LoadCommand { cmd, cmdsize }))
 }
 
+fn parse_dylib(input: &[u8]) -> IResult<&[u8], DylibObject> {
```
TommYDeeee, line 899: Can you please add docstring comment?

yara-x/src/modules/macho/mod.rs
```diff
@@ -839,6 +896,37 @@ fn parse_load_command(input: &[u8]) -> IResult<&[u8], LoadCommand> {
...
+        DylibObject {
+            name: name.into(),
+            timestamp,
+            compatibility_version,
+            current_version,
+        },
+    ))
+}
+
+fn parse_dylib_command(input: &[u8]) -> IResult<&[u8], DylibCommand> {
```
TommYDeeee, line 922: also comment please :)

yara-x/src/tests/mod.rs
```diff
@@ -2874,6 +2874,111 @@ fn test_macho_module() {
         0x00, 0x00, 0x00, 0x00, // align
     ];
 
+    let macho_x86_64_dylib_file = [
```
TommYDeeee, line 2877: Just a quick note, this is not a request. I am not sure if it is the best to store this kind of big real file-like representations for testing purposes here. Right now I am working on a test framework that will take something like this as an input and compares whole protobuf output with desired output. So this will became unnecessary. 
But in case we still want to have tests like this here I would move at least this into another file (similarly like in [C YARA](https://github.com/VirusTotal/yara/blob/master/tests/blob.h). If we also choose this approach I would add also more test cases and test files, that would also cover not just `dylib` addition.

TommYDeeee:
Thank you for this addition! I have left few suggestions. Don't take them as granted, just something to think about

yara-x/src/tests/mod.rs
latonis: @TommYDeeee I created a blob.rs file in tests/testdata, open to how we feel about this, but it seemed to make sense if we're going to have blobs of data instead of the macho itself

TommYDeeee: Thank you for this, it LGTM.

JFYI: I am working on test framework that will eventually make this obsolete. Everything should be described in this https://github.com/VirusTotal/yara-x/pull/44 PR.
 But TLDR: There will be a built-in test framework that will take multiple intel HEX representations of binary files (similarly like you already did) and it will compare the module output as a whole with expected output. So everything will be compared, not just few cases like we are doing now. The only thing I know for now that will not or can't be compared is exported functions and those should at least for be tested individually: https://github.com/VirusTotal/yara-x/blob/107275cef8e43fc553a835685e9adaef6ee819c7/yara-x/src/tests/mod.rs#L2960
  But again thank you for this anyway as it is always good to know that it is tested and working 👍

latonis: Happy to refactor when it becomes obsolete with the new framework for sure. 😸 thanks for taking a look at all of this

TommYDeeee: Don't worry. I will do it as a part of bigger restructure, tomorrow I plan to add output testing and after that I will move all test files from C YARA and also add everything from here. You can check it afterwards.

TommYDeeee: Just to let you know, I have finished the test framework, but also realised I don't want to make this PR dependant on mine or vice-versa. So once either https://github.com/VirusTotal/yara-x/pull/44 or this PR will be merged I will rebase the other and add rework the tests for dylib module.

latonis: sounds good @TommYDeeee, if i can help in anyway let me know 😄

latonis: Saw the testing changes were merged to main, i've gone ahead and refactored the testing to conform with it and it's aligned with main now @TommYDeeee

yara-x/src/tests/mod.rs
```diff
@@ -3099,7 +3100,7 @@ fn test_macho_module() {
             macho.dylibs[1].timestamp == 2
     }
     "#,
-        &MACHO_X86_64_DYLIB_FILE
+        &dylib_data
```
TommYDeeee, line 3103: Same for all bellow:
There is probably no need to have this also here as it is just double testing the same fields that are already tested in `test_modules` framework. The only things that can't be tested there and is better to include here are:

- constants
- exported functions

Sure there is no problem having it also here, but in long-term scenario it will just increase the code and places to maintain in case something changes (instead of just regenerating goldenfiles).

latonis: ah ok I see now, the duplicate tests have been removed :smile:

TommYDeeee (approved):
LGTM! Thanks for this addition

plusvic (approved):

PR #48, opened 2023-10-19
feat: underscores in numeric literals

### LINKED ISSUE #14: Allow using underscores (_) as part of numeric literals
Allow using underscores (_) as part of numeric literals like in [Rust](https://doc.rust-lang.org/rust-by-example/primitives/literals.html). 

You could write:
  * `1_000_000` instead of `1000000`
  * `10_KB` instead of `10KB`
  * `0xcafe_babe` instead of `0xcafebabe`

...etc

latonis:
Work to implement and close #14

- modified `pest` grammar to allow for underscores following at least one number in a numeric literal, including floating point and those ending with the size suffixes
- added tests as well

Also tested with the following rule
```yara
rule test {
  condition:
    2_000 == 2000 and 100KB == 1_00KB and 0o12 == 1_0 and 0x2_1 == 0x21 and 0x31_1 == 7_8_5
}
```


latonis:
this is no longer relevant with the parser rewrite i believe, closing

PR #53, opened 2023-11-22
feat: YAML serializer

TommYDeeee:
Dumping module for YARA-X. This module can be run by `yr dump <binary-file> <options>`, where options are not required and can be one of: `--modules/-m=<module>` and `--output-format/-o=<format>`. 
Currently supported output formats are:
- json
- yaml (human readable)

Supported modules are for now ELF, MACH-O, PE and LNK, other modules can be easily added in future.
This module takes binary file and dumps parsed information to STDOUT. Modules used for parsing can be either selected and forced by using `--modules/-m=<module>`  option or if left empty it is automatically. User can specify also multiple modules used for parsing and output of all modules will be shown. Format can be selected via `--output-format/-o=<format>`. If left empty `human-readable YAML` format is automatically selected. This format is basically valid YAML with additional colors and comments. This format also supports additional integer representation which can be selected by protobuf field descriptor and is also described in `Module Developer's Guide.md`.

Together with this I have also added parametrized tests for end2end testing and did a minor change in `macho` module, where dylib version was represented as an integer. It would make much more sense to have this as a string (last two digits represents patch version, previous two minor version and rest is major version number). 

Integer field representation descriptors were added only into `macho` module. I think module author should do both while developing the module so feel free to add it into other modules that produces output or change it in existing modules. Colors for different parts of message can be easily changed in `ColorsConfig` structure. Indentation is also specified with constant that can be changed. Right now I have selected 4 spaces as indentation to match the JSON output with both colors and indentation.


TommYDeeee:
Reworked implementation of: #50

PR #56, opened 2023-11-25
feat: min_version load commands for macho module

latonis:
Implemented `min_version_command` load_command parsing for each OS/device.
- LC_VERSION_MIN_MACOSX
- LC_VERSION_MIN_IPHONEOS
- LC_VERSION_MIN_WATCHOS
- LC_VERSION_MIN_TVOS


latonis:
Thoughts on adding a string representation of the nibble encoded version and sdk values, @plusvic or @TommYDeeee? 

Example with the test binary: 
```
 version: 657664 -> version_str: "10.9.0"
 sdk: 657920 -> sdk_str: "10.10.0"
```
###

latonis:
Decided to implement the string versions of the sdk and version, as I'd want them if I were writing rules for these values. :smile_cat:

TommYDeeee:
LGTM

> Decided to implement the string versions of the sdk and version, as I'd want them if I were writing rules for these values
I would also stick with the string values for this kind of fields. As this is something that was not in C YARA we don't need to maintain backwards compatibility and it is much more readable for user to use in rules or get in module output. This is also something that [otool](https://www.unix.com/man-page/osx/1/otool/) does for Mach-O files parsing.

latonis:
Cleaned up the code a little bit with that nice version helper function by @TommYDeeee :smile_cat:, this one is ready for review whenever @plusvic. Happy New Year! :heart:

latonis:
Gonna mark this one as Draft as I'd prefer #67 to be reviewed and merged first.

latonis:
Rolled this into the refactor for the parser implementation for all of my Mach-O work in #78.

PR #67, opened 2024-01-03
feat: mach-o load command parsing features + fix bug

latonis:
- implemented LC_ID_DYLINKER, LC_LOAD_DYLINKER, and LC_DYLD_ENVIRONMENT load commands
- fixed bug in parsing for RPath and Dylibs where swap needed to occur before taking offset
- implemented search functions for rpath and dylibs with `macho.rpath_present` and `macho.dylib_present`

### DIFF
--- a/yara-x/src/modules/macho/mod.rs
+++ b/yara-x/src/modules/macho/mod.rs
@@ -1111,7 +1111,7 @@ fn parse_rpath_command(
         swap_rpath_command(&mut rp);
     }
 
-    let (input, path) = take(cmdsize - offset)(input)?;
+    let (input, path) = take(rp.cmdsize - rp.offset)(input)?;
 
     rp.path = path.into();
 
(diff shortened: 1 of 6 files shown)


yara-x/src/modules/macho/mod.rs
```diff
@@ -1599,27 +1693,29 @@ fn handle_dylib_command(
     size: usize,
     macho_file: &mut File,
 ) -> Result<(), MachoError> {
-    if size < std::mem::size_of::<DylibCommand>() {
+    // 24 bytes for all integer fields and offset in Dylib/DylibCommand
+    // fat pointer of vec makes for inaccurate count
+    if size < 24 {
+        dbg!("here, oopsie");
```
TommYDeeee, line 1699: I would remove this or at least change the message otherwise LGTM

TommYDeeee (approved):

yara-x/src/modules/macho/mod.rs
latonis: ah clippy didn't alert me of that, ty, removing now!

plusvic (approved):

PR #65, opened 2023-12-31
feat: LC_UUID parsing for Mach-O binaries

latonis:
Implemented `LC_UUID` parsing for Mach-O binaries
- relevant tests added
-

PR #68, opened 2024-01-07
feat: parse Mach-O build_version load commands

latonis:
- parsing build version load commands
- added another Mach-O test file as none were multi-arch and contained LC_BUILD_VERSION


latonis:
Converting to draft until others have been approved and merged, as the proto definitions need to be updated each time.
#56 -> #65 -> #67

latonis:
This has been rolled into #78 with the refactor.

PR #72, opened 2024-01-13
feat: adds --test flag in yr fmt

shellcromancer:
Adds -t/--test flag in yr fmt which returns an error code if formatting made changes on the file.

This feature exists in a few other language formatting tools ([jsonnet](https://github.com/google/go-jsonnet/blob/master/cmd/jsonnetfmt/cmd.go#L47-L48), [terraform](https://developer.hashicorp.com/terraform/cli/commands/fmt#check), [gofmt](https://github.com/golang/go/blob/master/src/cmd/gofmt/gofmt.go#L33)) which is useful as part of stylistic CI/CD checks

### DIFF
--- a/yara-x-cli/src/commands/fmt.rs
+++ b/yara-x-cli/src/commands/fmt.rs
@@ -33,6 +33,8 @@ pub fn exec_fmt(args: &ArgMatches) -> anyhow::Result<()> {
     let formatter = Formatter::new();
 
     if let Some(files) = files {
+        let mut changed_files = Vec::new();
+
         for file in files {
             let input = fs::read(file.as_path())?;
 
@@ -44,7 +46,7 @@ pub fn exec_fmt(args: &ArgMatches) -> anyhow::Result<()> {
             let output = formatted.into_inner();
 
             if *test.unwrap() && input != output {
-                process::exit(3)
+                changed_files.push(file.display().to_string());
             }
 
             if *write.unwrap() {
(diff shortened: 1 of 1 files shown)


shellcromancer:
Signed the Google CLA, should be ready for a maintainer review, thanks! 👍

yara-x-cli/src/commands/fmt.rs
```diff
@@ -18,27 +19,38 @@ pub fn fmt() -> Command {
...
 
-                File::create(file.as_path())?
-                    .write_all(formatted.into_inner().as_slice())?;
+            formatter.format(input.as_slice(), &mut formatted)?;
+            formatted.rewind()?;
+
+            let output = formatted.into_inner();
+
+            if *test.unwrap() && input != output {
+                process::exit(3)
```
plusvic, line 47: Why not `process::exit(2)`? Exit code 2 is not used for any purpose yet. 

I'm also wondering if it wouldn't be better to return the path of the files that requires changes, as `gofmt -l` does. This provides more information to the user. If you call `yr fm` with multiple files, you may want to know which files are changed and which don't.

shellcromancer: Thank you for taking a look! I updated this to use exit code 2, and output a list of files that require changes. 

image

plusvic:
In https://github.com/VirusTotal/yara-x/commit/4b0889938df1feff849e6c9cdd4d84ab59eeec86 I made multiple changes to the `fmt` command. Now it always exits with error code 0 if files didn't change or 1 if otherwise.

shellcromancer:
Makes sense, thanks for including the feature in your changes. 👍

PR #77, opened 2024-01-31
fix: add support for FAT Mach-O files in yr dump auto module selection

TommYDeeee:
Automatic module selection for `yr dump` command did not take into consideration, that FAT Mach-O files do not have `magic` field in top level and instead of it they have `fat_magic`. This PR fixes it and automatic module selection produces output for valid FAT Mach-O files without having to manually specify which module should it use.


TommYDeeee:
Addressed in: https://github.com/VirusTotal/yara-x/pull/76

PR #78, opened 2024-02-02
feat: parsing code signature data + more load commands for mach-o

latonis:
This PR implements `LC_CODE_SIGNATURE` parsing for Mach-O and then parses the CS_Superblob, CS_Blob, and CS_Indexes that are available via the offsets in the code signature load command. 
- available to parse in the CS_blobs are certificates and entitlements
- both implemented

Additionally implemented `LC_UUID`, `LC_DYLD_INFO`, `LC_DYLD_INFO_ONLY`, `LC_BUILD_VERSION`, `LC_VERSION_MIN_MACOSX`, `LC_VERSION_MIN_IPHONEOS`, `LC_VERSION_MIN_TVOS`,  and `LC_VERSION_MIN_WATCHOS`


latonis:
`modules::pe::tests::checksum` is failing on the `nightly` test but passing on the other such as `stable` and `no-default-features` 🤔

latonis:
@plusvic, everything (including all the old PRs for Mach-O) has been refactored for the new parsing style and is ready for review 😸

lib/Cargo.toml
```diff
@@ -188,6 +188,8 @@ yara-x-parser = { workspace = true }
 yara-x-proto = { workspace = true }
 
 lingua = { version = "1.6.0", optional = true, default-features = false, features = ["english", "german", "french", "spanish"] }
+roxmltree = "0.19.0"
```
plusvic, line 191: Both `roxmltree` and `cryptographic-message-syntax` crates are only used by the `macho` module, they should be made an optional dependency that is included only if the `macho-module` feature is enabled.

lib/src/modules/macho/mod.rs
```diff
@@ -182,6 +182,42 @@ fn ep_for_arch_subtype(
     None
 }
 
+/// The function for checking if any dylib name present in the main Mach-O or embedded Mach-O files
+/// contain a dylib with the desired name
+///
+/// # Arguments
+///
+/// * `ctx`: A mutable reference to the scanning context.
```
plusvic, line 190: You don't need to document these functions with that level of details, with an "arguments"  and "returns" sections. Particularly the `ctx` shouldn't be included in the description.

The documentation for this function should be written from the perspective of someone using the function in YARA. The idea is that I can copy and paste the documentation for these exported functions and put it in the public documentation for the module. The `ctx` argument is an internal argument that is not visible from YARA.

Use the PE module as model of how to write the documentation for exported functions: https://github.com/VirusTotal/yara-x/blob/main/lib/src/modules/pe/mod.rs#L192-L203

latonis: ack, thanks Victor will rework the doc comments for the functions 😄

lib/Cargo.toml
latonis: Fixed i believe 😸

lib/src/modules/macho/mod.rs
latonis: updated all the functions i've added 😸

lib/src/modules/protos/macho.proto
```diff
@@ -10,13 +10,63 @@ option (yara.module_options) = {
   rust_module: "macho"
 };
 
+message MinVersion {
+  optional uint32 device = 1;
```
plusvic, line 14: If values in this field are one of the values in the `DEVICE_TYPE` enum, make this field of type `DEVICE_TYPE`. This way it will be shown in the YAML output as `device: MACOSX` instead of `device: 36`.

latonis: Fixed 😄

plusvic:
@latonis I had to remove part of the code introduced in this PR, specifically the portion related to signature parsing (see: https://github.com/VirusTotal/yara-x/commit/154994e9e1cb1c24a8e08313dca9740f97a8fa27). The problem was the `cryptographic-message-syntax` crate, which introduced a lot of dependencies on other crates that don't make too much sense. This dependencies were creating linking issues when I tried to compile a larger Golang program that uses YARA-X under the hood. For instance, the `cryptographic-message-syntax` depends on `reqwest`, which in turns depends on `tokio` and some other large crates.

I did a quick try to replace `cryptographic-message-syntax` with the `cms` crate, but didn't found a direct equivalent to the `SignedData::parse_ber` function. If you can take a look at it, it will be great.

latonis:
> @latonis I had to remove part of the code introduced in this PR, specifically the portion related to signature parsing (see: https://github.com/VirusTotal/yara-x/commit/154994e9e1cb1c24a8e08313dca9740f97a8fa27). The problem was the `cryptographic-message-syntax` crate, which introduced a lot of dependencies on other crates that don't make too much sense. This dependencies were creating linking issues when I tried to compile a larger Golang program that uses YARA-X under the hood. For instance, the `cryptographic-message-syntax` depends on `reqwest`, which in turns depends on `tokio` and some other large crates.
> 
> I did a quick try to replace `cryptographic-message-syntax` with the `cms` crate, but didn't found a direct equivalent to the `SignedData::parse_ber` function. If you can take a look at it, it will be great.
> 
> 

Ack, I'll take a look. The unfortunate part (which you discovered) is that Apple uses BER format for the embedded certificates and this was the only library I could find that supports parsing them. All others supported DER but not BER. I noticed the `cryptographic-message-syntax' crate has an issue for this actually (all the dependencies): https://github.com/indygreg/cryptography-rs/issues/9

I'll see if it's worth implementing a PR to make the nuisance (and optional) dependencies truly optional with crate features or if implementing the parsing myself is better.

latonis:
I've opened a PR on the lib to implement optional dependencies via features: 
https://github.com/indygreg/cryptography-rs/pull/21

latonis:
Ah okay, I'll take a look into that. 😸

latonis:
Coming back to this, Apple previously used BER encoding and has now switched to DER format [1]. So we will need to support both to support the scanning and parsing of older binaries and malware. I'll get a PR going to parse DER format (the newer one), but I'll need to continue investigating how we can parse the BER format ones to parse older samples. 

[1] https://developer.apple.com/documentation/xcode/using-the-latest-code-signature-format

PR #90, opened 2024-03-06
feat: implement mach-o signed certificate parsing

latonis:
@plusvic, created this PR to show what the deps will look like from `cryptographic-message-syntax` with the proper features disabled. Currently, I am using my branch at https://github.com/latonis/cryptography-rs/tree/cms-features as I am not sure when https://github.com/indygreg/cryptography-rs/pull/21 will be approved and/or merged in. 

I dug into RustCrypto, but unfortunately they deprecated BER format decoding when they deprecated the `pkcs7` crate (see: https://docs.rs/pkcs7/latest/pkcs7/). As such, RustCrypto doesn't provide a mechanism for parsing the BER format signed data. The RFC spec for this (see https://datatracker.ietf.org/doc/html/rfc5652) allows for BER format, but it seems like Apple is the outlier here and predominantly using it.

Without the ability to parse BER format, we lose the ability to parse Mach-O signature data


latonis:
> According to [RustCrypto/formats#779](https://github.com/RustCrypto/formats/issues/779) and [RustCrypto/formats#1321](https://github.com/RustCrypto/formats/pull/1321) they intend to implement BER decoding in the `cms` crate, precisely for addressing issues with parsing Apple stuff.

Ah ok, I didn't find those when perusing! Thanks, I'll keep an eye on it.

latonis:
In that case, I will hold off and close this PR. I do plan on shifting from the XML entitlements in mach-o to parsing the new DER encoded requirements if available as documented in https://developer.apple.com/documentation/xcode/using-the-latest-code-signature-format.

PR #93, opened 2024-03-21
feat: implement dylib and entitlement hashing for macho

latonis:
Implemented a macho similarity function `dylib_hash()`  and `entitlement_hash()` which is similar to `imphash` or any other attribute hashing mechanism. 

This will hash dylib entries as defined in https://github.com/g-les/macho_similarity/blob/main/implementation.md#dylib-hashing.
- I verified the hashing is consistent in both implementation and output by using `python ~/src/macho_similarity/ktool_macho_bulk_hashing.py -f <macho>`

I plan on implementing symtab_hash, import_hash, and export_hash in future PRs once I get each parsed out as well.


latonis:
i will get this PR fixed up this weekend :)

latonis:
Rust nightly channel is throwing up some weird compilation errors. Maybe a discussion for if nightly should be gated for the commits?

plusvic:
> Rust nightly channel is throwing up some weird compilation errors. Maybe a discussion for if nightly should be gated for the commits?
> 
> @plusvic, thoughts on nightly not being a gated check for tests?

I agree, we could run tests with nightly once a day, just like we run the coverage tests.

latonis:
@plusvic, this one is ready for review, the import and export parsing relies on me writing the trie parsing which requires uleb128 parsing among other things.

I think it would be good to get these two in for detection opportunities while I get the imports and exports parsed :)

lib/src/modules/macho/mod.rs
```diff
@@ -264,6 +266,71 @@ fn has_rpath(ctx: &ScanContext, rpath: RuntimeString) -> Option<bool> {
...
+
+    // we need to check again as the nested file dylibs could be empty too
+    if dylibs_to_hash.is_empty() {
+        return None;
+    }
+
+    let dylibs_to_hash: String = dylibs_to_hash
+        .iter()
+        .map(|d| {
+            std::string::String::from_utf8(d.name.clone().unwrap())
```
plusvic, line 289: If `d.name` is guaranteed to be `Some`, why make it optional?

latonis: The generated Macho for the `macho.proto` keeps those fields as options even when the `macho.proto` marks them as `required` I believe.

latonis: image
image

PR #117, opened 2024-05-21
build: cargo: Set default release profile to use LTO

jpds:
Linux distros use this target for their end-user packages, as such this needs to be the default for those users to take advantage of the performance benefits.


plusvic:
I'm not convinced about making LTO the default because compilation is *very* slow, specially when used together with `codegen-units = 1`. The default release profile is good enough, and produces very fast code. The real benefit I see in using the `release-lto` profile is reducing the executable size.

I prefer using as the default release configuration whatever `cargo` choses as default. Anything that deviates from `cargo`'s default, is a special profile used for specific purposes.

PR #121, opened 2024-05-22
feat: Implement multiline string literals in metadata.

wxsBSD:
This commit implements multiline string literals in metadata section.

```
rule a {
  meta:
    a = """
I'm a multiline string literal!

Hooray!

\"test\"

I also handle escapes, \x41\x42\x43!

... and emojis 🤖!
"""
  condition:
    true
}
```

While I'm here, also quote the metadata string values when printing the ast.


wxsBSD:
I'm not sure this is ready to merge yet because of an oversight. Basically, strings and multiline strings are treated the same internally so we have no way of knowing which it is which means the formatted doesn't know how to properly format them. I'll see if I can fix this quick.

Issue #120, opened 2024-05-22
Rule metadata C-API


codedninja:
Sorry about opening another issue about this, I am not as knowledgeable with Rust as I thought. I tired to implement the C-API for rule's metadata but it hasn't gone as well as I have hoped. Thank you in advance!

Would it be possible to request the rule metadata be added to the C-API as well?

plusvic:
Implemented in bfa57afde38bae9c9dae6828b5865f21b70948d1

Issue #118, opened 2024-05-21
Rule's meta field


codedninja:
The parser parses the meta fields but the compiler doesn't do anything with those fields. I was curious if there is any plans to add that functionality in anytime soon.

I utilize meta fields for additional information when processing a match in my code.

Fryyyyy:
+1 to outputting this when there's a match, please!

jpds:
Added in https://github.com/VirusTotal/yara-x/commit/1e816a70471a32313d7194ebc199aff3abab87c6

plusvic:
I've exposed the metadata both in the Rust and Python APIs, but I would like to hear more about your use cases. This is something that you want to use also from the CLI?

`yara` has a `--print-meta` options that prints the metadata information for every matching rule. But I don't think that option is really useful if you want to use that information in some script or automation. Instead of replicating the `--print-meta` option in `yara-x`, my idea is adding an option to the CLI that dumps the information about matches in JSON or YAML (or both). This way, the matching information can be easily parsed by other tools, and it can include not only the rule name for each match, but also the rule's metadata.

Would you find it helpful?

codedninja:
I am totally fine with it being in the APIs, I mainly use meta tags for giving the rule a score or sending alerts to email/webhooks/etc inside of the meta tags.

Thanks for adding the Meta tag support so quickly! In a few I will create a PR for adding support for to the go package.

Fryyyyy:
Will this be exposed in the Golang bindings as well?

PR #123, opened 2024-05-23
feat(go): expose rule's metadata in Go API

codedninja:
Reference issue: #120 & #118


plusvic:
Now that I see an implementation I'm reconsidering the use of JSON as a format for exchanging metadata information. There are some issues with JSON that make it far from ideal:

* We lose information about whether a number is actually a float or an integer.
* For non-UTF8 strings we are going to get an array of numbers, which will be translated to `[]float` instead of `[]byte`.

codedninja:
When I wrote the tests I noticed that as well, maybe a similar implementation as the original Yara did? Which is what I was originally aiming for when attempting to implement it in the C-API.

```C
struct YR_META
{
  DECLARE_REFERENCE(const char*, identifier);
  DECLARE_REFERENCE(const char*, string);

  int64_t integer;
  int32_t type;
  int32_t flags;
};
```

plusvic:
I've refactored the C API for obtaining the rule's metadata. See: 9f90eaab37a764cc90c40881fc60e448483bb7bf

codedninja:
There is a slight error with the `float` variant inside of the `YRX_METADATA_VALUE` union. When the header file gets generated it makes it into "two or more data types in declared specifier" due to `float` being a reserved keyword.

While you decide on the new name for the float variant, I shall temporary rename to get the go package working.

plusvic:
> There is a slight error with the `float` variant inside of the `YRX_METADATA_VALUE` union. When the header file gets generated it makes it into "two or more data types in declared specifier" due to `float` being a reserved keyword.
> 
> While you decide on the new name for the float variant, I shall temporary rename to get the go package working.

Fixed in 497c678cb288dbfdfc12415741e301d724735c97

PR #132, opened 2024-05-28
feat: implement mach-o export trie parsing and export hashing function

latonis:
Implements mach-o trie export parsing as well as the `export_hash()` function mentioned previously in #93. 

Once this is reviewed, I will begin work on import parsing and hashing :smile_cat:


lib/src/modules/macho/parser.rs
```diff
@@ -327,6 +333,22 @@ impl<'a> MachO<'a> {
                 // everything else
             };
         }
+
+        if let Some(ref dyld_info) = macho.dyld_info {
+            let offset = dyld_info.export_off as usize;
+            let size = dyld_info.export_size as usize;
+            if offset < data.len() {
+                let export_data = &data[offset..offset + size];
```
plusvic, line 341: `&data[offset..offset + size]` can panic because `size` is read from the file and it can be arbitrarily large

This should be enough, you don't even need to check for `offset < data.len()`:

```rust
if let Some(export_data) = data.get(offset..offset.saturating_add(size)) {
  ...
}
```

lib/src/modules/macho/parser.rs
```diff
@@ -888,6 +911,87 @@ impl<'a> MachOFile<'a> {
...
+    /// Parser that parses the exports at the offsets defined within LC_DYLD_INFO and LC_DYLD_INFO_ONLY
+    fn exports(
+        &mut self,
+    ) -> impl FnMut(&'a [u8]) -> IResult<&'a [u8], Vec<String>> + '_ {
+        move |data: &'a [u8]| {
+            let exports = Vec::<String>::new();
+            let Ok((remainder, _)) =
+                self.parse_export_node()(data, 0, BStr::new(""))
+            else {
+                todo!()
```
plusvic, line 989: Is this unfinished?

lib/src/modules/macho/parser.rs
latonis: ah good catch, thank you! fixed it in some other parts as well instead of the data.len() check!

latonis:
thanks for the quick review @plusvic! :smile: , fixed everything and should be ready :rocket:

PR #135, opened 2024-05-30
feat: mach-o import parsing and import hash function

latonis:
Parsing imports for Mach-O as well as the accompanying `import_hash()` function.


latonis:
I need to refactor the `sleb` parsing to mirror the simplification done in https://github.com/VirusTotal/yara-x/commit/1bd014d8f8eaf9a628c4efccd0c7c57d795c69f4 to the `uleb` parser.

Also, probably worth changing imports to `bytes` instead of `String` to avoid panics on imports that are not UTF-8.

plusvic (approved):

latonis:
I will fix `sleb` to be panic free like `uleb` has been as well as merge conflicts. Will then fuzz for a while to ensure everything is good 🙏

PR #161, opened 2024-07-23
feat: add json output to scan command.

wxsBSD:
When using scan command you can now output in json or json-pretty format. The former is just a single JSON string per line while the latter is a pretty-printed JSON object.

Due to the output subsystem you can do things like:

./yr scan --output-format json rules/test.yara /bin | jq .path

and it will work, but you won't see the information about currently scanning files or the "final" message which tells you the number of files scanned. I'm guessing this is because the superscan is aware that stdout is being redirected somehow? Either way, this is a nice feature because it means we get the ability to pipe the JSON output to things like jq easily.

### DIFF
--- a/cli/src/commands/scan.rs
+++ b/cli/src/commands/scan.rs
@@ -6,7 +6,7 @@ use std::sync::Mutex;
 use std::time::{Duration, Instant};
 
 use anyhow::{bail, Context, Error};
-use clap::{arg, value_parser, ArgAction, ArgMatches, Command};
+use clap::{arg, value_parser, ArgAction, ArgMatches, Command, ValueEnum};
 use crossbeam::channel::Sender;
 use superconsole::style::Stylize;
 use superconsole::{Component, Line, Lines, Span};
@@ -19,6 +19,15 @@ use crate::commands::{
 };
 use crate::walk::Message;
 use crate::{help, walk};
+
+#[derive(Clone, ValueEnum)]
+enum OutputFormats {
+    Text,
+    Json,
+    JsonPretty,
+}
+
+const STRINGS_LIMIT: usize = 120;
 
 #[rustfmt::skip]
 pub fn scan() -> Command {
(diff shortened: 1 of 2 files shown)


wxsBSD:
Good point, if we are going to output JSON it should be valid JSON. I'll fix it by just making things a list of the current objects.

wxsBSD:
This is actually harder than I initially thought it would be. I think the problem is that I need access to the output channel upon initialization, but only once to add in the `[` character at the start. The init function is called per-thread. I would then need to know when the walker is done (I considered adding a `finalize` function to the walker so I could use it to output the `]` but that felt wrong.

I was investigating the approach of just storing all the JSON objects in a vector in `state` in the walker and then printing them when it is done, but the output channel is gone by then, so that's not going to work. Also, it feels wrong to store all the results when using one of the json outputs when the design is clearly to give responsive output as the scan is happening.

There is also another problem where a trailing comma in a JSON list is illegal (at least according to python, `json.loads('["a", "b", "c",]')` will throw an error). So something like `[{"path": ... }, {"path": ...},]` wouldn't work. I would need to know when the last thread is done scanning so I can emit a json record with no trailing comma, which is rather difficult to do here too.

What if I made it clear in the help (and docs, once I figure out where those are =b) that it isn't a single JSON output but rather one JSON object per file scanned. For regular json output this means one JSON object per line which is easy enough to parse in your scripting language of choice, and for the json-pretty option it will only work with things that are expecting multiple JSON objects like jq does. In essence, if you want to parse it with a custom script you will want to use regular json output and parse each line as a separate object, and use `json-pretty` for well, pretty printing only.

I've updated this PR with a refactor of the code which should make it a bit easier to read in case you have any suggestions.

plusvic:
Sure, I agree. Let's output a stream of JSON objects instead of a single array. It turns out that transforming the stream of objects into an array is very easy with `jq`.

```
yr scan --output-format=json <RULES> <DIRECTORY> | jq -s .
```

We can explain all this in the documentation.

PR #171, opened 2024-08-03
feat: Implement tag printing.

wxsBSD:
This commit adds a -g argument to the scan command which will print the tags of matching rules.

One thing I realized (and this is also true for metadata) is that if there is no tags we are still including them in the JSON (just as an empty list). When we print in text mode we have checks if the tags are empty and skipping them if so.

I'm open to changing the JSON output so that we do not include a "meta" and "tags" key if they are empty.


wxsBSD:
I also added in tag filtering. It currently works like `-t` does in original YARA.

PR #174, opened 2024-08-07
feat: Add -c options.

wxsBSD:
Add the -c option to output the number of rules that matched on each file. This is also implemented for the JSON output options.

I took this opportunity to refactor the code slightly to make the closure in the walker a little smaller by moving the logic on what to output and how into a separate function.

### DIFF
--- a/cli/src/commands/scan.rs
+++ b/cli/src/commands/scan.rs
@@ -77,6 +77,10 @@ pub fn scan() -> Command {
                 .help("Print rule tags")
         )
         .arg(
+            arg!(-c --"count")
+                .help("Print only number of matches")
+        )
+        .arg(
             arg!(--"disable-console-logs")
                 .help("Disable printing console log messages")
         )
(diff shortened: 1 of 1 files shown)


cli/src/commands/scan.rs
```diff
@@ -641,6 +629,57 @@ impl ScanState {
...
+        let mut matching_rules = scan_results.non_matching_rules();
+        let match_count = matching_rules.len();
+        print_matching_rules(args, &file_path, &mut matching_rules, output);
+        return match_count;
+    } else {
+        let mut matching_rules = scan_results.matching_rules();
+        let match_count = matching_rules.len();
+        print_matching_rules(args, &file_path, &mut matching_rules, output);
+        return match_count;
+    };
```
plusvic, line 660: In this implementation `--negate` can't be use used in conjunction with `--count`, you either use one or the other. However, I don't see a reason for that, I think that they should be allowed to be used together. Also, when you use `--negate`, the number returned by `--count` should be the number of non-matching rules. In other words, `--negate` should invert the results, when used without `--count` it returns the non-matching rules, and when used with `--count` it should return the number of non-matching rules.

wxsBSD: Thanks! I was trying to mimic the behavior of original YARA here, but I like you're suggestion better.

PR #179, opened 2024-08-16
feat: Add "modules" subcommand to display modules available.

wxsBSD:
This adds a "developers" feature that when enabled will turn on the "debug" command - and without it the command is not available. Prior to this the command was always available but hidden.

I think the ability to list modules is a nice feature for users to have available, so I'm not sure I like the idea of putting this behind a features flag that is off by default. Does it make sense to have "debug" be behind developers but move the ability to list modules to it's own command that is on by default?


cli/Cargo.toml
```diff
@@ -22,6 +22,10 @@ path = "src/main.rs"
 # to work fine in Linux.
 profiling = ["pprof"]
 
+# Enable the "debug" command. Calling it "developer" here because calling it
+# "debug" is confusing as that implies a debug build.
+developer = []
```
plusvic, line 27: I would name it `debug-cmd`. This way the purpose of this feature is clearer.

lib/src/modules/mod.rs
```diff
@@ -256,4 +256,12 @@ pub mod mods {
         info.lnk = protobuf::MessageField(invoke::<Lnk>(data));
         info
     }
+
+    /// A vector of all module names. Useful for displaying currently compiled
+    /// modules.
+    ///
+    /// See the "debug modules" command.
+    pub fn module_names() -> Vec<&'static str> {
+        super::BUILTIN_MODULES.keys().map(|&k| k).collect()
```
plusvic, line 265: Return an iterator instead of a vector, so that allocating the vector is not always required.

wxsBSD: Done. I was trying to avoid it because the return type now exposes the Module type, which means I had to alter it's visibility. It didn't feel right to do that, so please let me know if there is a better way!

wxsBSD: Ah, I think I am starting to understand more. Thanks!

PR #183, opened 2024-08-30
feat: support for more complex argument in module main fn

chudicek:
The pre-existing code did not support the additional input (the -x <FILE> flag input) necessary for some modules. Hence a modification of the relevant part of the code was made to enable passing a more complex struct to the `#[module_main]` function (as opposed to just a single `_: &[u8]` input argument).

The implementation itself should be in line with what was discussed with @TommYDeeee


lib/src/scanner/mod.rs
```diff
@@ -346,6 +374,25 @@ impl<'r> Scanner<'r> {
             ScannedData::Mmap(mapped_file)
         };
 
+        Ok(data)
+    }
+
+    /// Scans a file.
+    pub fn scan_file<'a, 'target, 'meta>(
+        &'a mut self,
+        scan_input: &ScanInput<'target, 'meta>,
```
plusvic, line 383: This is a backward incompatible change in the API that is not required nor desirable. In the vast majority of cases the API user just wants to scan a file and don't want to be bothered with additional details like the metadata. Moreover, if we would pass the metadata to this function, why not simply add a new argument? Like this:

```rust
pub fn scan_file<P: AsRef<Path>, M: AsRef<Path>>(&self, path: P, meta: Option<M>)
```

The introduction of  the`ScanInput` struct to the API seems unnecessary. However, this is not the most problematic issue with this design. The main problem is that the additional metadata is not always the same for every module. The `-x` argument in `yara` has the following form: `--module-data=MODULE=FILE`, where `MODULE` is a the module name and `FILE` the path to the file passed to that module. This option can be used multiple times, passing different files to each module.  You can have modules `foo` and `bar`, each expecting a different kind of metadata and invoke YARA like:

```shell
yara -module-data=foo=file1.json --module-data=bar=file2.json ...
```

The current design doesn't cover this case, all modules will see exactly the same metadata.

My proposal is adding a `set_module_meta` to the `Scanner` object, which works in a similar way to [set_module_output](https://docs.rs/yara-x/latest/yara_x/struct.Scanner.html#method.set_module_output).

Before calling `scan`, the user calls `set_module_meta` in order to specify the metadata that will be passed to a specific module. This function will receive both the module's name and a slice with the metadata for that module. Once `scan` is called, the metadata for that module is cleared and in subsequent calls to `scan` the module won't receive any metadata unless the user calls `set_module_input` again before calling `scan`.

Of course, this requires changing the signature for the module's main function, but instead of introducing a new struct `ScanInputRaw` it could simply be changed to: 

```
main(data: &[u8], meta: Option<&[u8]>)
```

lib/src/modules/mod.rs
```diff
@@ -210,10 +215,13 @@ pub mod mods {
...
     /// # use yara_x;
-    /// # let data = &[];
-    /// let elf_info = yara_x::mods::invoke::<yara_x::mods::ELF>(data);
+    /// # use yara_x::ScanInputRaw;
+    /// # let data = ScanInputRaw { target: &[], meta: None };
+    /// let elf_info = yara_x::mods::invoke::<yara_x::mods::ELF>(&data);
     /// ```
-    pub fn invoke<T: protobuf::MessageFull>(data: &[u8]) -> Option<Box<T>> {
+    pub fn invoke<T: protobuf::MessageFull>(
+        data: &ScanInputRaw,
```
plusvic, line 223: Instead of changing the signature of `invoke`, add a new function:

 ```rust
pub fn invoke_with_meta(data: &[u8], meta: &[u8])
```

Besides backward-compatibility (which is not that important at this point when the API is still beta), the reason is usability. Most users of this API only need to call `invoke` without any metadata at all, an the API should facilitate the most common use case.

File: lib/src/scanner/mod.rs (hunk 1 above)
Comment (chudicek) [reply lib/src/scanner/mod.rs]: thx for the feedback

implemented in [b9a6140](https://github.com/VirusTotal/yara-x/pull/183/commits/b9a61401036315f758fb1ba4138e5085b68de42f)

File: lib/src/modules/mod.rs (hunk 2 above)
Comment (chudicek) [reply lib/src/modules/mod.rs]: implemented in [6718dd4](https://github.com/VirusTotal/yara-x/pull/183/commits/6718dd41f6da680fced617714db2d3afb1f397c2)

File: cli/src/commands/scan.rs [hunk 3]
```diff
@@ -290,8 +322,14 @@ pub fn exec_scan(args: &ArgMatches) -> anyhow::Result<()> {
                 .unwrap()
                 .push((file_path.clone(), now));
 
+            let target_file = file_path.as_path();
```
plusvic, line 325: This line can be removed.

lib/build.rs
```diff
@@ -40,6 +40,11 @@ fn generate_module_files(proto_files: Vec<FileDescriptorProto>) {
         }
     }
 
+    // Sort the modules by name to avoid non-deterministic builds
+    modules.sort_by_key(|it| it.3.to_owned());
```
plusvic, line 44: Already done in https://github.com/VirusTotal/yara-x/commit/4a9e0cc19a14c40e0a73e0f6f04559c172cd57e1

lib/src/modules/mod.rs
```diff
@@ -225,6 +231,7 @@ pub mod mods {
     /// dispatch version of the structure returned by the YARA module.
     pub fn invoke_dyn<T: protobuf::MessageFull>(
         data: &[u8],
+        meta: Option<&[u8]>,
```
plusvic, line 234: We have `invoke` and `invoke_dyn`, and now we are adding `invoke_with_meta`. For consistency we should have `invoke_with_meta_dyn` instead of changing the signature for `invoke_dyn`.

lib/src/modules/mod.rs
```diff
@@ -244,16 +251,17 @@ pub mod mods {
     /// # Example
     /// ```rust
     /// # use yara_x;
-    /// # let data = &[];
-    /// let modules_output = yara_x::mods::invoke_all(data);
+    /// let modules_output = yara_x::mods::invoke_all(&[], None);
     /// ```
-    pub fn invoke_all(data: &[u8]) -> Box<Modules> {
+    pub fn invoke_all(data: &[u8], meta: Option<&[u8]>) -> Box<Modules> {
```
plusvic, line 256: This is tricky, all modules are invoked with the same metadata and that doesn't make sense. I would leave `invoke_all` untouched for now. None of the invoked modules expect metadata anyways.

py/src/lib.rs
```diff
@@ -357,7 +357,7 @@ impl Scanner {
         Python::with_gil(|py| {
             scan_results_to_py(
                 py,
-                self.inner.scan_file(path).map_err(map_scan_err)?,
+                self.inner.scan_file(path.as_path()).map_err(map_scan_err)?,
```
plusvic, line 360: Why is this change necessary? The `scan_file` API is not changing, right?

lib/src/scanner/mod.rs
```diff
@@ -457,6 +499,26 @@ impl<'r> Scanner<'r> {
...
     }
 
+    /// Updates the metadata for a module specified by its fully-qualified name.
+    ///
+    /// If the `meta` argument is `None`, the metadata for the module is removed.
+    ///
+    /// See [`Scanner::module_meta`] for the reasoning behind choosing `Arc<_>`
+    pub fn set_module_meta(
+        &mut self,
+        module_full_name: &str,
```
plusvic, line 509: For simplicity this API should receive the module name ("pe", "elf", "macho", etc). The concept of fully-qualified name an internal concept, and most API users won't know what is this. It's ok if the API can work with both names, but if only one of them is accepted, it should be short name.

lib/src/scanner/mod.rs
```diff
@@ -507,6 +499,26 @@ impl<'r> Scanner<'r> {
...
+        meta: Option<&Arc<[u8]>>,
```
plusvic, line 510: I believe it's possible to pass `Option<&[u8]>` here. It's will be harder due to lifetime issues, but that's actually the desired API. The scanner doesn't really need to retain the metadata after the scan has finished, therefore a reference to the metadata should be enough. However, in case lifetimes makes this hard to do, I think that using `Rc` is enough. I don't see a reason for using `Arc` as the scanner is not `Send` nor `Sync` anyways. You are supposed to use one scanner per thread, and I don't see why anyone would need to pass the same metadata to multiple scanners running in different threads.

We can use `Rc` for now and try to change it to `&[u8]` in a future commit.

plusvic: After thinking about this a little bit more, I believe I was wrong. Passing `&[u8]` to `set_module_meta` may be impossible without unsafe code. However, there is an alternative design that makes possible to get rid of those smart pointers `Arc` and `Rc` from the API. Suppose that we have a type like this:

```rust
pub struct ScanOptions<'a> {
}

impl<'a> ScanOptions<'a> {
    fn set_metadata(&mut self, module_name: &str, meta: &'a [u8]) {
    }
}
```

A new function would be added to `Scanner`:

```rust
    pub fn scan_with_options<'a>(
        &'a mut self,
        data: &'a [u8],
        options: ScanOptions<'a>,
    ) -> Result<ScanResults<'a, 'r>, ScanError> {
       ...
    }
```

Then you can use it as:

```rust

let mut scanner = Scanner::new(rules)
let mut options = ScanOptions::default();

options.set_metadata("pe", pe_metadata));
options.set_metadata("elf", elf_metadata);

scanner.scan_with_options(data, options)
```

I think this makes everything simpler because you don't need to store the modules metadata in `Scanner` nor in `ScannerContext`, and you don't have a lifetimes nightmare. It also opens the door for adding more scan options in the future by adding more methods to `ScanOptions`.

cli/src/commands/scan.rs
chudicek: fixed in [cca5adf](https://github.com/VirusTotal/yara-x/pull/183/commits/cca5adfaeb3671482a47f5b0a1a8e3f7d2a70fe8)

lib/build.rs
chudicek: removed in [cca5adf](https://github.com/VirusTotal/yara-x/pull/183/commits/cca5adfaeb3671482a47f5b0a1a8e3f7d2a70fe8)

lib/src/modules/mod.rs
chudicek: done in [cca5adf](https://github.com/VirusTotal/yara-x/pull/183/commits/cca5adfaeb3671482a47f5b0a1a8e3f7d2a70fe8)

lib/src/modules/mod.rs
chudicek: youre right, this does not make sense the moment we add the possibility for each module to have its metadata

removed in [cca5adf](https://github.com/VirusTotal/yara-x/pull/183/commits/cca5adfaeb3671482a47f5b0a1a8e3f7d2a70fe8)

py/src/lib.rs
chudicek: removed in [cca5adf](https://github.com/VirusTotal/yara-x/pull/183/commits/cca5adfaeb3671482a47f5b0a1a8e3f7d2a70fe8)

lib/src/scanner/mod.rs
chudicek: good catch, the name remained after a version of the code that used the full module name to identify the modules. this was later changed, the name was not

fixed in [cca5adf](https://github.com/VirusTotal/yara-x/pull/183/commits/cca5adfaeb3671482a47f5b0a1a8e3f7d2a70fe8)

lib/src/scanner/mod.rs
chudicek: mb for not checking scanner already is `!Send + !Sync`, should have been `Rc<_>`

rewrote it according to your suggestions anyway in [b699aa0](https://github.com/VirusTotal/yara-x/pull/183/commits/b699aa0d18894e047c23839d2423a85869fbad11)

lib/src/scanner/mod.rs
```diff
@@ -349,17 +388,24 @@ impl<'r> Scanner<'r> {
     where
         P: AsRef<Path>,
     {
-        let target = Self::load_file(target.as_ref())?;
+        self.scan_file_with_options(target, ScanOptions::new())
```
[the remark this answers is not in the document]
chudicek: done in [1ed2ec7](https://github.com/VirusTotal/yara-x/pull/183/commits/1ed2ec79edb6ef01f628c291b96e5285cd0a229d)

cli/src/commands/dump.rs
```diff
@@ -22,6 +22,7 @@ enum SupportedModules {
     Elf,
     Pe,
     Dotnet,
+    Eml,
```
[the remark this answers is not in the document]
chudicek: yep, should not be there

PR #190, opened 2024-09-05
fix: add a minimal-default feature

qjerome:
Attempt at making stabler and simpler the exclusion of `parallel-compilation` feature without changing default behavior.

- stabler: end-user doesn't need to track changes in `yara-x` default features to reflect it into his own project on update
- simple: only one feature to use in `yara-x` dependency instead of several

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -159,13 +159,13 @@ text-module = [
 # conditions of a rule to check against other epoch time.
 time-module = []
 
-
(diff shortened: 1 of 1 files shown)


plusvic (approved):

plusvic:
Not needed anymore, `parallel-compilation` was removed from default features in https://github.com/VirusTotal/yara-x/commit/831db9c21972672916269ed7272cf1dd3abdc65e.

PR #197, opened 2024-09-16
feat: implement `with` statement

TommYDeeee:
This expression would allow us to declare a local variable that will be used in remainder of a corresponding section. 

```
condition: 
    with foo = <expr>, 
         bar = <expr> : (
       <a boolean expression where you can use identifiers foo and bar>
    )
```

Another advantage of this approach is that we can use it inside a loop.

```
for all offset in (10,20,30) : (
   with val = uint8(offset) | uint8(offset + 4) | uint8(offset + 8) : (
      val == 0x10000 or 
      val == 0x20000 or 
      val == 0x40000
   )
)
```

This implementation follows [RFC](https://github.com/VirusTotal/yara/discussions/1783) proposal.


plusvic (commented):
Great job!

parser/src/parser/mod.rs
```diff
@@ -1582,6 +1583,45 @@ impl<'src> ParserImpl<'src> {
...
+    }
+
+    /// Parses `with` identifiers.
+    ///
+    /// ```text
+    /// WITH_IDENTIFIERS :=
+    ///     IDENT `=` EXPR (`,` IDENT `=` EXPR)*
+    ///
+    fn with_identifier(&mut self) -> &mut Self {
+        self.begin(WITH_IDENTIFIERS)
```
plusvic, line 1613: I think `WITH_IDENTIFIERS` is not the best name for this. Something like `WITH_DECLARATIONS` is more appropiate, or `WITH_DECLS` for coherency with other names.

parser/src/parser/tests/testdata/with-2.cst
```diff
@@ -0,0 +1,104 @@
...
+      CONDITION_KW@13..22 "condition"
+      COLON@22..23 ":"
+      NEWLINE@23..24 "\n"
+      WHITESPACE@24..26 "\t\t"
+      BOOLEAN_EXPR@26..160
+        BOOLEAN_TERM@26..160
+          WITH_EXPR@26..160
+            WITH_KW@26..30 "with"
+            WHITESPACE@30..31 " "
+            WITH_IDENTIFIERS@31..90
```
plusvic, line 20: The `WITH_IDENTIFIERS` rule contains a sequence of comma-separated declarations in the form `identifier = expr`, however I miss some grammar rule that contains each individual declarations. Something like:

``
WITH_IDENTIFIERS :=  WITH_IDENTIFIER [, WITH_IDENTIFIER]*
``

But as mentioned before, I would change change the names of these rules as: 

```
WITH_DECLS :=  WITH_DECL [`,` WITH_DECL]*

WITH_DECL := IDENT `=` EXPR
```

lib/src/tests/mod.rs
```diff
@@ -514,6 +514,69 @@ fn for_in() {
...
+                foo = test_proto2.array_int64[i],
+                bar = test_proto2.array_int64[i + 1] : 
+                    (
+                        foo == 1 and bar == 10
+                    )
+        )"#
+    );
+
+    #[cfg(feature = "test_proto2-module")]
+    condition_false!(
```
plusvic, line 568: I miss a test where the expression is a function call, and another where it is a structure:

```
with foo = some_func() : ( ... )
```

```
with foo = test_proto2.nested : (  foo.nested_int64_one == 1 )
```

lib/src/compiler/ir/mod.rs
```diff
@@ -610,6 +613,14 @@ pub(in crate::compiler) struct ForIn {
     pub stack_frame: VarStackFrame,
 }
 
+/// A `with` expression (e.g `with $a, $b : (..)`)
+#[derive(Debug)]
+pub(in crate::compiler) struct With {
+    pub identifiers: Vec<Var>,
```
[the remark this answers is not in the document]
TommYDeeee: Yes that sounds good! Done in https://github.com/VirusTotal/yara-x/pull/197/commits/3858aafe1452bba33aa2334bc6da55ced10a15f9

parser/src/ast/mod.rs
```diff
@@ -364,6 +364,22 @@ pub enum OfItems<'src> {
     BoolExprTuple(Vec<Expr<'src>>),
 }
 
+/// A `with` expression (e.g `with $a : (..)`)
```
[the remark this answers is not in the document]
TommYDeeee: you are right this is a typo, fixed in: https://github.com/VirusTotal/yara-x/pull/197/commits/3858aafe1452bba33aa2334bc6da55ced10a15f9

parser/src/ast/mod.rs
```diff
@@ -576,6 +592,9 @@ pub enum Expr<'src> {
 
     /// A `for <quantifier> <vars> in ...` expression. (e.g. `for all i in (1..100) : ( ... )`)
     ForIn(Box<ForIn<'src>>),
+
+    /// A `with` expression (e.g. `with $a : ( ... )`)
```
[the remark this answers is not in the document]
TommYDeeee: same as above

parser/src/parser/mod.rs
TommYDeeee: I have unified it across the code under `declarations`, it should be all good now

parser/src/parser/tests/testdata/with-2.cst
TommYDeeee: Added in: https://github.com/VirusTotal/yara-x/pull/197/commits/3858aafe1452bba33aa2334bc6da55ced10a15f9

lib/src/tests/mod.rs
TommYDeeee: You are right, I have added those cases to cover everything. Thank you

TommYDeeee:
> Great job!

Thank you for a review. I have incorporated all the requested changes in: https://github.com/VirusTotal/yara-x/pull/197/commits/3858aafe1452bba33aa2334bc6da55ced10a15f9

hopefully it is all good now.

plusvic (approved):

Issue #192, opened 2024-09-09
Ideas to improve "yr fmt"


wxsBSD:
I haven't dug into it in detail yet but as discussed elsewhere, we should add various options to allow people to control the formatting (and even other things, like variable names, maybe?) in "yr fmt" - I'm happy to do the work here but want to open this issue to get input on it.

For example, I'm thinking of things like

- [x] Use tabs vs spaces
- [x] Controllable number of spaces for identation
- [x] Curly brace on new line for rule declaration (eg: `rule foo {\n` vs `rule foo\n{`)
- [x] Newline after section headers
- [ ] How loops should be formatted (eg: `for any i in (0..1): (\n` vs `for any i in (0..1):\n(`)
- [x] Optional indentation of rule body
- [x] Optional alignment of metadata and patterns
- [ ] Alphabetize metadata
- [ ] Alphabetize strings
- [x] Remove trailing whitespace
- [x] Enforce certain metadata fields (e.g. they must exist and be of type x)
- [ ] Line breaks between strings using different prefixes, e.g.

I'm sure there are plenty more here to be found, especially as I dig into the code, but hoping to collect your thoughts @plusvic (and others) on ways to improve "yr fmt" so I can go off and start implementing.

EDIT: I'm putting the things mentioned below into the list above so I can edit it to mark the ones I have done.

plusvic:
I like those and add a few more:

- [ ] Optional indentation of rule body:

```
rule foo {
strings:
  $a = ...
condition:
  $a
}
```

vs

```
rule foo {
  strings:
    $a = ...
  condition:
    $a
}
```

- [ ] Optional alignment of metadata and patterns.

```
rule foo {
  meta: 
    short = "..."
    very_long = "..."
  strings:
    $short = "..."
    $very_long = "..."
  condition:
    all of them
}
```

vs

```
rule foo {
  meta: 
    short     = "..."
    very_long = "..."
  strings:
    $short     = "..."
    $very_long = "..."
  condition:
    all of them
}
```

My plan was having a TOML configuration file not only for the code formatter, but for the CLI as a whole, where code formatting is a section. This crate could be useful for finding the user's home directory: https://crates.io/crates/home

wxsBSD:
[...]

PR #205, opened 2024-09-24
feat: Implement toml config for fmt command.

wxsBSD:
This implements a config option which can be used to set options for "yr fmt".

An example config file is included as "wxsfmt.toml" which currently has everything disabled just to demonstrate the behavior.

If any option is not set it will fall back to the default values provided by the "fmt" command which currently match the defaults provided by the library.


wxsBSD:
This is not ready to be merged yet, but it is at a point where the overall design should be discussed.

The idea is to have defaults that can be overridden with the config file, which is toml format. An example config file is:

```toml
[rule]
indent_section_headers = false
indent_section_contents = false

[meta]
align_values = false

[patterns]
align_values = false
```

I think it is worth breaking up the config file into sections, which will make it easier to write and document. Currently there are three sections:

- `rule` for anything that applies globally to all rules. This is currently things like indenting the section headers (`strings:`, `meta:`, etc)
- `meta` for anything related to metadata contents. Currently only `align_values` is supported here.
- `patterns` for anything related to `strings` section. Currently only `align_values` is supported here.

The idea is to expose the rest of the options in the formatter through the config file, but if we choose to change the format of the config file now is a good time to do that.

I'm open to changing names on any of these if you have any opinions on that.

plusvic:
I like the approach, and the `figment` create looks like a great choice, but I would make the configuration file a global one that not only affects the `fmt` command. The idea is using the same config file for other settings not related to `fmt` in the future. As such, the file should have a `[fmt]` section that contains the rest of the subsections you propose.

I would also like that users don't need to specify the config file in the command-line (that option can remain), but the program should look for well-known locations looking for a `yara-x.toml` with the configuration.

wxsBSD:
OK, I've added support for loading it from `${HOME}/.yara-x.toml` and also moved things around a bit so that config related structures and loading are in their own module. The main `yr` binary now reads the config file and passes the `fmt` section to the `fmt` command, which is something we can easily extend to the other commands as the config file grows support for them.

Turns out the config file format I gave above is total garbage (I misread some stackoverflow post) and this is a much better way to do it:

```
# Config file for YARA-X. For full documentation see XXX (INSERT DOCS URL HERE)

[fmt]
rule.pants = false # Invalid keys are ignored.
rule.indent_section_headers = true
rule.indent_section_contents = false
meta.align_values = false
patterns.align_values = false
```

The above is my current `.yara-x.toml` file, which demonstrates what I have in mind for how this can be extended. Each command can have it's own section and we can express values using the dot notation if further nesting is necessary. If no config file is found it will use the defaults defined in `lib/config/mod.rs`

If you think this config format works well I'll update this PR with the actual documentation and then I can go about tackling the rest of the options in discussed in #192.

plusvic:
I like the latest changes.

fmt/src/tokens/mod.rs
```diff
@@ -212,6 +212,8 @@ pub(crate) enum Token<'a> {
     // Non-control tokens
     //
     Whitespace,
+    #[allow(dead_code)]
```
wxsBSD, line 215: @plusvic I'm not sure this is the right way to do this.

PR #224, opened 2024-10-12
feat: More fmt improvements

wxsBSD:
This adds support for optional newline before curly brace in the rule declaration.

```
// This is the default
rule a {
```

vs

```
rule a
{
```

It also adds support for empty lines before and after section headers. Each one is a separate option - you can have them before section headers but not after, vice versa, both, or none.

```
rule a {

  meta:

    foo = "bar"
```

vs

```
// This is the default (no empty lines)
rule a {
  meta:
    foo = "bar"
```

This also fixes an issue where we were not inserting newlines before meta definitions, so if you had two meta definitions on the same line the formatter left them like that. They are now fixed so that each meta definition goes on a separate line.


wxsBSD:
This is more work for #192 - if this is merged I'll follow-up with tests that handle all the options added so far. The formatter is starting to get more complicated with these options and I want to make sure we have good test coverage to help ensure proper functionality as I continue to work on the other requested features.

plusvic (approved):

plusvic:
A test case is broken.

This unformatted code:
```
rule test {
meta:

  a = 1

strings:
  $a = "foo"

condition:

  true}
```

Was expected to be:
```
rule test {
  meta:
    a = 1

  strings:
    $a = "foo"

  condition:

    true
}
```

I think that by default the actual result after formatting should be:
```
rule test {
  meta:
    a = 1

  strings:
    $a = "foo"

  condition:
    true
}
```

So the existing code missed to remove the newline after `condition:`.

wxsBSD:
Sorry, just read your comment. I fixed the test but it isn't what you want. I'll change the defaults and fix the tests once I'm done traveling today.

wxsBSD:
OK, I think this should be good now. The default for `empty_lines_before_section_headers` is now `true` and it apples to all section headers EXCEPT for the first one.

PR #230, opened 2024-10-26
test: Add tests for fmt config options.

wxsBSD:
This commit adds tests for the configurable formatter. Each configuration option is tested individually, as they each should be independent of each other. If we have a need to test multiple configuration options simultaneously in the future this setup should be able to do it easily.

While here I also found a bug in how tab stops were treated by the formatter when removing trailing whitespaces. After indentation was performed the step to remove trailing whitespace was not properly checking for the inserted tab characters.

There was also another bug with handling the emptyline before the section header which I fixed.

Also, fix a bug in the docs where the `meta.align_values` and `meta.align_patterns` options were incorrectly set. They changed to true in a previous commit but I did not update the docs then.


wxsBSD:
I moved a lot of tests around for this, but I think it makes sense to separate out the "default" tests and the ones were we are testing configuration options for the formatter. Let me know if you don't like this change and I'll figure something out.

plusvic (approved):

PR #235, opened 2024-11-03
feat: mach-o certificate parsing implemented

latonis:
With https://github.com/indygreg/cryptography-rs/pull/21 being merged in now, we can have a much smaller dependency tree without `reqwest` and similar that were causing problems that caused the revert in https://github.com/VirusTotal/yara-x/commit/154994e9e1cb1c24a8e08313dca9740f97a8fa27 for parsing certificates in mach-o binaries.

With not much progress for BER encoded certificates being made in https://github.com/RustCrypto/formats/issues/779, I felt this was worth adding for the capability to search certificates in mach-o 😸


plusvic:
I still have concerns about introducing another set of dependencies that implement an ASN parser. Depending on different sets of crates that implement basically the same are far from ideal.

I don't recall which was the issue we were trying to solve here, but I think that in the `der_parser` crate we already have the tools needed for parsing BER encoded certificates, as long we implement the required ASN structures by ourselves.

latonis:
> I still have concerns about introducing another set of dependencies that implement an ASN parser. Depending on different sets of crates that implement basically the same are far from ideal.
> 
> I don't recall which was the issue we were trying to solve here, but I think that in the `der_parser` crate we already have the tools needed for parsing BER encoded certificates, as long we implement the required ASN structures by ourselves.

Yeah, that's a fair concern. I can take a look at using `der_parser` and implementing the needed parts myself. 😄

PR #239, opened 2024-11-07
feat: json output

chudicek:
added the `json` variant to the existing `--output-format` arg, including the integration with the `--print-meta`, `--print-strings`, `--print-tags`, `--count`, `--tag` args

refactored the `ndjson` output implementation to match the `json` style

refactored the existing code to use DI (via dynamic dispatch) to separate the output logic from the rest

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -57,7 +57,7 @@ log = { workspace = true, optional = true }
 protobuf = { workspace = true }
 protobuf-json-mapping = { workspace = true }
 serde_json = { workspace = true, features = ["preserve_order"] }
-serde = { workspace = true }
+serde = { workspace = true, features = ["derive"] }
 yansi = { workspace = true }
 yara-x = { workspace = true, features = ["parallel-compilation"] }
 yara-x-parser = { workspace = true }
--- a/cli/src/commands/check.rs
+++ b/cli/src/commands/check.rs
@@ -133,6 +133,8 @@ pub fn exec_check(args: &ArgMatches) -> anyhow::Result<()> {
         },
         // Finalization
         |_, _| {},
+        // Walk done
+        |_| {},
         // Error handling
         |err, output| {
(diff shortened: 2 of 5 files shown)


plusvic:
I'm reviewing this more carefully and there's one thing that seems wrong. There's a field named `rulesets` that is an array of dictionaries with two fields: `file` and `timestamp`, but they contain information about the matched files, not about the files containing the rulesets. What's the intention with that `rulesets` field?

Also, when used with the `--count` option, the number of matches is printed to the console, but not as part of the JSON object. Mixing JSON output with non-JSON output is problematic.

chudicek:
Good catch, the `rulesets` field should indeed hold the information about the files that contain the rules. (Looked into how to get those for a few minutes, but did not find a straightforward way to do so yet.)

As for the `--count` option - the idea was to maintain backwards compatibility, since our fork of the original `yara` used this output format. Will contact someone who can answer whether we can change it to be a field in the json (iirc we discussed that and it should not be a huge problem).

plusvic:
In 8baf1944f48d0aad36cd720e622d77bf77416719 I've done some refactoring in the code and simplified some things.  What I did was basically make the `json` output to have the same format as `ndjson` but all at once at the end of the scan. I've removed the `rulesets` and `version` fields but I'm open to make changes if you need it.

plusvic:
I agree on changing the format for the meta section, using an object at the top level, and including version information. The only thing that I have doubts about its the inclusion of the list of rulesets in the JSON. 

In cases where you pass pre-compiled rules to the CLI that information won't be around. Also, maybe in the future I add some support for ignoring bad rules in a source file and continue with the compilation without failing, in such cases you can't assume that all the files that were passed to the compiler were included in the the compiled rules, rendering this information inaccurate. I feel that this a very specific feature that may be useful in your use-case, but it's unlikely to be used by anyone else.

PR #248, opened 2024-11-19
feat: implement sym_hash() function for mach-o module

latonis:
Implement the `sym_hash()` function for the mach-o module to hash all of the symbol table entries, if the binary has not been stripped. This did not include new parsing, only utiIizing the entries we already parsed
- tests included


lib/src/modules/macho/mod.rs
```diff
@@ -446,6 +446,39 @@ fn import_hash(ctx: &mut ScanContext) -> Option<RuntimeString> {
...
+
+    // we need to check again as the nested symbol table could be empty too
+    if symtab_to_hash.is_empty() {
+        return None;
+    }
+
+    let symtab_hash_str: String = symtab_to_hash
+        .iter()
+        .filter_map(|e| {
+            Some(String::from_utf8(e.to_vec()).unwrap().trim().to_lowercase())
```
plusvic, line 470: `String::from_utf8` could fail if the symtab contains a non UTF-8 string. This is probably the case in malformed files that could make the program crash.

lib/src/modules/macho/mod.rs
```diff
@@ -446,6 +446,39 @@ fn import_hash(ctx: &mut ScanContext) -> Option<RuntimeString> {
     Some(RuntimeString::new(digest))
 }
 
+/// Returns an md5 hash of the symbol table in the mach-o binary
+#[module_export]
+fn sym_hash(ctx: &mut ScanContext) -> Option<RuntimeString> {
+    let macho = ctx.module_output::<Macho>()?;
+    let mut md5_hash = Md5::new();
```
plusvic, line 453: Allocate the `Md5` object after the initial validation. In cases were the file doesn't have a symtab the allocation is not needed.

plusvic:
Clippy says...

```
error: this `.filter_map` can be written more simply using `.map`
   --> lib/src/modules/macho/mod.rs:467:35
    |
467 |       let symtab_hash_str: String = symtab_to_hash
    |  ___________________________________^
468 | |         .iter()
469 | |         .filter_map(|e| {
470 | |             Some(String::from_utf8(e.to_vec()).unwrap().trim().to_lowercase())
471 | |         })
    | |__________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_filter_map
    = note: `-D clippy::unnecessary-filter-map` implied by `-D clippy::all`
    = help: to override `-D clippy::all` add `#[allow(clippy::unnecessary_filter_map)]`
```

lib/src/modules/macho/mod.rs
plusvic: Notice that you can use the `BStr` type for wrapping a `&[u8]`. This type provides `to_lowercase`, `trim` and `join`, but it works with non-UTF-8 strings.

lib/src/modules/macho/mod.rs
latonis: Fixed 😄

latonis:
Addressed the review and comments @plusvic 😸 , thanks for the quick review!

lib/src/modules/macho/mod.rs
```diff
@@ -446,6 +446,38 @@ fn import_hash(ctx: &mut ScanContext) -> Option<RuntimeString> {
...
+    // we need to check again as the nested symbol table could be empty too
+    if symtab_to_hash.is_empty() {
+        return None;
+    }
+
+    let mut md5_hash: digest::core_api::CoreWrapper<md5::Md5Core> = Md5::new();
+
+    let symtab_hash_entries = symtab_to_hash
+        .iter()
+        .map(|e| e.trim().to_lowercase().as_bstr().to_owned())
```
plusvic, line 470: Here it's more efficient to do `BString::new(e.trim().to_lowercase())`. `to_lowercase` returns a `Vec<u8>`, when you do `.as_bstr().to_owned()`, you are basically cloning the vector and wrapping it in a `BString`. By using `BString::new` you are given the ownership of the original vector to `BString`, and no clone is performed.

latonis: Ahhh, I see now. Thanks for pointing that out Victor, appreciated 😄

PR #249, opened 2024-11-20
feat(capi+go): expose match data (strings) via YRX_MATCH

egibs:
I noticed that the API currently does not expose the strings from a given match while the CLI does: https://github.com/VirusTotal/yara-x/blob/ba8c764f8490fb2037fae539a22e0981e72188fc/cli/src/commands/scan.rs#L851

Since matches are already plumbed into the Go API, this was a pretty small lift but I still wanted to capture the desired behavior with a couple of test updates.

If there's a better or more appropriate implementation for this, please let me know!

cc: @tstromberg

### DIFF
--- a/capi/include/yara_x.h
+++ b/capi/include/yara_x.h
@@ -107,6 +107,10 @@ typedef struct YRX_BUFFER {
 
 // Contains information about a pattern match.
 typedef struct YRX_MATCH {
+  // Slice of the data of the match.
+  const uint8_t *data;
+  // Length of the data slice.
+  size_t data_len;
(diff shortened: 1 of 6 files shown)


plusvic:
Not including the matched data in the structure was an intentional decision. The matching strings can be many, and very long, and exposing them in Go API requires copying the data via `C.GoBytes`. Making copies of very long strings is expensive.

Also, these copies are not necessary after all, because the user of the API already knows their lengths and offsets within the original data, and the user of the API must have access to the data in order to pass it to the scanner in the first place.

PR #254, opened 2024-11-27
feat: Implement extra data from dump mode.

wxsBSD:
This commit is a prototype for adding a "dump" mode for modules. This is a flag that can be passed to modules that tells them to perform "extra" processing that they would normally not do during "scan" mode. This is meant for things like `pe.imphash()` which is normally calculated on-demand.

### DIFF
--- a/cli/src/commands/dump.rs
+++ b/cli/src/commands/dump.rs
@@ -89,7 +89,7 @@ pub fn exec_dump(args: &ArgMatches) -> anyhow::Result<()> {
         stdin().read_to_end(&mut buffer)?
     };
 
-    let mut module_output = invoke_all(&buffer);
+    let mut module_output = invoke_all_dump(&buffer);
 
     if let Some(modules) = requested_modules {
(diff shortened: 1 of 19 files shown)


plusvic:
I initially created this issue without giving much thought to its implementation. However, after considering it further, it seems more complex than I first anticipated.

I'm not entirely convinced by the current approach, as it involves adding new fields to the protobuf. These fields feel somewhat redundant and would also expand the API surface in a way that doesn’t feel natural. My original idea was to iterate through the list of WASM-exported functions, identify those exported by modules and suitable for invocation, call those functions, and inject their results into the output.

The challenge I overlooked is that the protobuf defines the output structure, and there’s no way to introduce additional fields without modifying the protobuf definitions. At this point, I’m uncertain if it’s feasible to achieve this without adding artificial fields to the protobuf. If that’s the case, I’d prefer leaving it as is.

wxsBSD:
Sounds good to me. I'll close this PR but leave the branch around in case anyone wants it.

PR #276, opened 2024-12-28
feat: mach-o certificate parsing implemented with der_parser

latonis:
As discussed in #235, implementing the certificate parsing with `der_parser` to reduce dependencies and potential headaches.

Depends on #275, as the ASN.1 parsing functions were specific to PE before this PR.


latonis:
Open to discussing how much CMS data we return.

Currently we expose the following:

```rust
Certificate {
      issuer: c.x509.issuer.to_string(),
      subject: c.x509.subject.to_string(),
      is_self_signed: c.x509.issuer
          == c.x509.subject,
  }
```

plusvic:
> Open to discussing how much CMS data we return.
> 
> Currently we expose the following:
> 
> ```rust
> Certificate {
>       issuer: c.x509.issuer.to_string(),
>       subject: c.x509.subject.to_string(),
>       is_self_signed: c.x509.issuer
>           == c.x509.subject,
>   }
> ```

I would like to expose as much data as in the `pe` module if possible. This is the structure that we are currently exposing in the `pe` module:

```rust
message Certificate {
  optional string issuer = 1;
  optional string subject = 2;
  optional string thumbprint = 3;
  optional int64 version = 4;
  optional string algorithm = 5;
  optional string algorithm_oid = 6;
  optional string serial = 7;
  optional int64 not_before = 8 [(yaml.field).fmt = "t"];
  optional int64 not_after = 9 [(yaml.field).fmt = "t"];
}
```

I guess this information is also available for Mach-O certificates, right? I would also add the `is_self_signed` field to the PE module.

PR #277, opened 2024-12-30
feat: implement the Formatter API for python lib

latonis:
- exposed the `format` API via the Python lib
- updated the typing stub
- documented the functions in Py
- added `__pycache__` to be ignored
- test for python formatting

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ macros/target
 site/node_modules
 site/public
 site/resources
+__pycache__
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -134,9 +134,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.94"
+version = "1.0.95"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c1fd03a028ef38ba2276dce7e33fcd6369c158a1bca17946c4b1b701891c1ff7"
+checksum = "34ac096ce696dc2fcabef30516bb13c0a68a11d30131d3df6f04711467681b04"
 
 [[package]]
 name = "arbitrary"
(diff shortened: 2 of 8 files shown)


plusvic:
I think the Python API for the formatter should mimic the Rust API as much as possible. For instance, the `Formatter::format` method should receive two file objects, one for reading the unformatted code and the other for writing the formatted code. This API allows using the `Formatter::format` method with in-memory data, but also with file objects, which is what most people will probably want, as they are formatting files in most cases. Additionally, this way the Python API doesn't need to make copies of the input or output data, it simply passes the file objects down to the `yara-x-fmt` API.

The `serialize_into` and `deserialize_from` methods can be used as examples of how to convert a Python's file object into a Rust reader or writer, that can be passed to the Rust API.

https://github.com/VirusTotal/yara-x/blob/4d590177b162832a155b1aec96158fb16192471b/py/src/lib.rs#L528-L548

latonis:
> I think the Python API for the formatter should mimic the Rust API as much as possible. For instance, the `Formatter::format` method should receive two file objects, one for reading the unformatted code and the other for writing the formatted code. This API allows using the `Formatter::format` method with in-memory data, but also with file objects, which is what most people will probably want, as they are formatting files in most cases. Additionally, this way the Python API doesn't need to make copies of the input or output data, it simply passes the file objects down to the `yara-x-fmt` API.
> 
> The `serialize_into` and `deserialize_from` methods can be used as examples of how to convert a Python's file object into a Rust reader or writer, that can be passed to the Rust API.
> 
> https://github.com/VirusTotal/yara-x/blob/4d590177b162832a155b1aec96158fb16192471b/py/src/lib.rs#L528-L548

Good call out 😸 , fixed in 53bae8ff

PR #281, opened 2025-01-07
refactor: json output vol2

chudicek:
# Json Output Redo

This PR contains changes in the json output format of the scan command (`--output-format=json`), as (mostly) discussed in the <https://github.com/VirusTotal/yara-x/pull/239> PR.

### `meta` field

The meta field is now an object, rather than an array of arrays.

Old format:

```json
"meta": [
	[
		"meta_1",
		"value_1"
	]
]
```

New format:

```json
"meta": {
	"meta_1": "value_1"
}
```

This maps more appropriately to the idea of the meta fields being a key-value collection.

### Top-level json

The top-level json is now formed from an object that then holds the hits as one of its fields, rather than the top-level json being an array of hits.

Old format:

```json
[
	{
		// hit object
	}
]
```

New format:

```json
{
	// some extra fields
	"hits": [
		{
			// hit object
		}
	]
}
```

The benefit of the new representation is that it allows us to include extra information about the scan itself, which is useful for archiving the outputs of the scans (see [version field](#version-field)).

### Version field

The top-level json holds a field with information about the used version of yara-x.

```json
{
	"version": "<the yara-x version (from Cargo.toml)>"
}
```

This is one of the possible extra pieces of information useful when archiving the output of the scan.

### Denormalized hit objects

The object containing info about a hit contains both info about the scanned file and the rule that it matched.

Old format:

```json
[
	{
		"path": "path to the scanned file",
		"rules": [
			{
				"identifier": "rule_that_matched"
				// ... meta, tags, strings
			}
		]
	}
]
```

New format:

```json
{
	"hits": [
		{
			"rule": "rule_that_matched",
			"file": "path to the scanned file"
			// ... meta, tags, strings
		}
	]
}
```

Reasoning behind this is backwards compatibility. This format is already relied on by our tooling. This is the format that we originally proposed in <https://github.com/VirusTotal/yara-x/pull/239/commits/00ccc34d660818211849976c0a20c11bca8f5fe0>.

---

Should there be any further questions about the reasoning behind any of the changes, I would be happy to discuss them in this PR.

### DIFF
--- a/cli/src/commands/scan.rs
+++ b/cli/src/commands/scan.rs
@@ -999,8 +999,6 @@ mod output_handler {
     #[derive(serde::Serialize)]
     struct OutputJson {
         version: String,
-        #[serde(skip_serializing_if = "Option::is_none")]
-        count: Option<usize>,
         hits: Vec<HitJson>,
     }
 
@@ -1144,7 +1142,6 @@ mod output_handler {
                 std::mem::take(&mut *lock)
             };
             let version = env!("CARGO_PKG_VERSION").to_string();
-            let count = self.output_options.count_only.then_some(hits.len());
 
             let rendered_json = match self.output_options.count_only {
                 true => {
@@ -1161,7 +1158,7 @@ mod output_handler {
                     serde_json::to_string_pretty(&json_output)
                 }
                 false => {
-                    let output_json = OutputJson { hits, version, count };
+                    let output_json = OutputJson { hits, version };
 
                     serde_json::to_string_pretty(&output_json)
                 }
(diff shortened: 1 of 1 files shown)


plusvic:
The only change that I would propose to this is renaming "hits" to "matches". Everywhere in the YARA's documentation we talk about matches instead of hits, so it can be a little bit confusing if we start using a different term in the JSON. Of course this will impact as your internal tooling is surely using the "hits" field. Are you ok with renaming this field even if it implies a change on your internal tools?

chudicek:
Checked with the involved team; it we are okay with using `"matches"` instead of `"hits"`

Changed in [0b38912](https://github.com/VirusTotal/yara-x/pull/281/commits/0b389125c45ebdc7dda9cedfcc80b4ebdba1516d)

PR #282, opened 2025-01-09
Implement basic linting

wxsBSD:
Apologies as this PR is a little bigger than I wanted, partially due to fixing a bug in config file handling and partially because it implements two linter checks (required metadata and rule name regexp).

# Config file handling improvements

Prior to this change if you had an invalid config file in your home dir we would just silently replace it with the default config, which is unexpected from a user standpoint (invalid configs should produce an error and refuse to run). To deal with this we now pass errors back up in the main command instead of using `.unwrap_or_default()`.

Also, rather than have each sub-command implement config file handling I moved it into main, to remove duplicate code. Also, because we are now propagating errors if you have an invalid config in `${HOME}/.yara-x.toml` I improved error messages so it is clear what file failed to parse:

```
Error: Unable to parse /Users/wxs/.yara-x.toml: invalid type: found signed int `0`, expected a boolean for key "default.fmt.patterns.align_values" in TOML source string
```

# Linting checks

To implement linting checks for required metadata and rule name format I chose to implement them in the compiler directly. I initially had an implementation with a stand-alone Linter type that worked on the AST from the parser. While this worked it turned out that getting it to use the existing Report/ReportBuilder mechanism was exceedingly intrusive as reporting is coupled too closely to the compiler (or at least I couldn't figure out how to make it work easily). Because of this I just went with implementing it directly in the compiler, which I think is also a good idea because we now get warnings anywhere the compiler is used, which can be quite beneficial for enforcing standards.

To use these linting checks I added a section in the config file and exposed the options for it there. See the config guide changes for details, but the TLDR is you can now specify a table (TOML term for a dictionary) where the key is the metadata identifier that is required and the value is the type that is required. Any rule that does not meet these requirements will get warnings from the compiler.

The other check is a regex that must match the rule name. Again, this is specified in the config file and warnings are issued by the compiler for any rule that does not match this regex.

The default value for both of these options is off so by default the compiler does not enforce any particular standards but you can opt in to it for linting if you want. The only place that currently sets these options is the `yr check` command.

If you're happy with these changes I'd like to also expose them in the various bindings so users can get lint warnings in the language of their choice.

### DIFF
--- a/docs/YARA-X Config Guide.md
+++ b/docs/YARA-X Config Guide.md
@@ -2,8 +2,8 @@ YARA-X Config Guide
 ===================
 
 YARA-X uses a configuration file for controlling the behavior of different
-commands. It currently supports the `fmt` command, but others will be added in
-the future.
+commands. It currently supports the `fmt` and `check` commands. More may be
+added in the future.
 
 The `yr` command looks in `${HOME}/.yara-x.toml` when starting up. If that file
 does not exist the default values are used.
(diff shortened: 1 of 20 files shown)


docs/YARA-X Config Guide.md
```diff
@@ -143,4 +145,48 @@ meta.align_values = true
...
+#
+# The key is the identifier and the value is the required type for that
+# identifier. Supported types are "string", "int", "float", "bool", "md5",
+# "sha1", "sha256", or "hash".
+#
+# To require that there be an "author" metadata field and the value must be a
+# string, and that there needs to be a date field and the value must be an
+# integer use this:
+#
+# meta = { author = "string", date = "int" }
```
xorhex, line 164: should meta be metadata in this case?

plusvic:
I wasn't convinced about the proposed API for a number of reasons:

- A new method was added to the `Compiler` for each type of check. If we add some other checks in the future this will bloat the compiler's API with multiple methods that are not really related to the compiler's main feature which is compiling.

- As each check is implemented directly in the compiler, the compiler's code is also more complicated, not only its API.

- Passing a `BTreeMap` type as part of the API exposes an implementation detail.

- The `yara_x` crate didn't look like a good place to put the `MetaValueType` type.

I've refactored the API to make it more flexible, adding a single method to the compiler (`Compiler::add_linter`). This method receives any type that implements the `Linter` trait. New types `linters::RuleNameMatches` and `linters::RequiredMetadata` implement this trait. This allows adding many more checks in the future without bloating the `Compiler` object. It also allows creating more complex linters with ergonomic APIs.

Issue #212, opened 2024-10-04
Inconsistent output formatting and unexpected symbols in `yr` scan command output


r0ny123:
The output formatting is inconsistent when running `yr` and `yara`. Specifically, `yr` prints the rule name followed by the full path of the matched file, which is quite lengthy and less readable, whereas `yara` only prints the filenames and subdirectories when the -r flag is supplied.

Additionally, the `yr` command output includes some unexpected `?` symbols in front of the path, which seems to be incorrect.

```
C:\Users\rony\MAL_FAMILY\SPARROWDOOR\LOADER>yr scan draculoader.yar .
generated_rule \\?\C:\Users\rony\MAL_FAMILY\SPARROWDOOR\LOADER\other\2531891691ef674345f098ef18b274091acdf3f2808cca753674599c043ccd7d

C:\Users\rony\MAL_FAMILY\SPARROWDOOR\LOADER>yara draculoader.yar . -r
generated_rule .\other\2531891691ef674345f098ef18b274091acdf3f2808cca753674599c043ccd7d
```

plusvic:
The prefix `\\?\` is still shown, but only when the target path passed to the CLI is an absolute path. This is a Windows thing, where absolute file paths are internally like that. The prefix could be removed, but it's not recommended. I'll keep it for now.

https://stackoverflow.com/questions/41233684/why-does-my-canonicalized-path-get-prefixed-with
https://users.rust-lang.org/t/understanding-windows-paths/58583

PR #283, opened 2025-01-09
fix: Workaround bug in globwalk.

### LINKED ISSUE #280: Panic with .\ directory prefix
When using YARA-X 0.12.0 a panic is generated if the path starts with `.\`. This is how PowerShell auto-completes folder names relative to the current directory. Example:

```
PS C:\<my_user_directory>\yara> yr scan -r "C:\<path>\rule.yar" .\samples\
thread '<unnamed>' panicked at C:\Users\runneradmin/.cargo\registry\src\index.crates.io-6f17d22bba15001f\globwalk-0.9.1\src\lib.rs:381:78:
called `Result::unwrap()` on an `Err` value: StripPrefixError(())
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Also affects Linux:

```
-> % ./yr scan -r rule.yar ./samples
thread '<unnamed>' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/globwalk-0.9.1/src/lib.rs:381:78:
called `Result::unwrap()` on an `Err` value: StripPrefixError(())
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Testing:
* `.\samples` - does not work
* `.\samples\` - does not work
* `samples\` - works
* `samples` - works
* `..\yara\samples\` - works

All the working examples are fine with the backslash substituted for a forward slash (on Windows), so the only difference is the `.\`.

wxsBSD:
As identified in #280 there is a bug in globwalk when using relative paths. Fix it by trying to canonicalize the path, and if that fails falls back to using the relative path. This is fine because the reason canonicalization failed is because the relative path was to a place that doesn't exist, and that does not trigger the bug (it errors out before we get there). We intentionally want this behavior so we get the nicer error reporting with colors. ;)

Fixes #280.

### DIFF
--- a/cli/src/commands/scan.rs
+++ b/cli/src/commands/scan.rs
@@ -1,6 +1,6 @@
 use std::borrow::Cow;
 use std::cmp::min;
-use std::fs::File;
+use std::fs::{canonicalize, File};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Mutex;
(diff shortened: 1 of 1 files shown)


plusvic:
This change has an undesired side-effect. When you pass a relative path to the scan command (like in `yr scan my_rule.yar .`), the paths of the matching files should be relative to that path (like `./foo/bar`), however, after this change all the paths reported by the CLI are absolute paths. That's specially problematic in Windows, where the absolute paths contain drive letters and the `\\?\` prefix. See #212.

plusvic:
Fixed in bccf19e0da145b33f373a54faa68e16635cd6f3a with a different approach.

PR #294, opened 2025-01-18
test: fix "cargo test" after cst change.

wxsBSD:
I apparently don't understand enough about rust to know why it is running test code from comments, but it apparently is.


plusvic:
> I apparently don't understand enough about rust to know why it is running test code from comments, but it apparently is.

Rust encourages you to write code examples that are also tests, precisely to prevent cases like this one, where the documentation gets out-of-date due to changes in the actual code.

PR #295, opened 2025-01-20
fix: reduce consumption of virtual address space.

### LINKED ISSUE #292: [Question] `mmap` `os error 12` in Cloud Run
I'm seeing an interesting panic when running Go code that leverages yara-x in Cloud Run when and scanning certain files:
Image

I imagine that this has something to do with Cloud Run's filesystem being entirely in-memory since I'm unable to reproduce this anywhere else outside of a Docker container using a tmpfs and memory limit. Additionally, even when using 24Gi+ of memory for the Cloud Run Service, creating a single scanner will result in this panic when scanning said files.

For reference, the scanner is being created with roughly ~15,000 rules, but it doesn't seem like that should require a reservation of 8GB of memory (`0x200000000` bytes) unless I'm completely misunderstanding.

https://github.com/VirusTotal/yara-x/blob/6a99e6ad15b288d868b6dd5421c812cadda7d1cd/lib/src/scanner/mod.rs#L217-L267

Here's the rough outline of what's happening:
```go
yxc, err := yarax.NewCompiler()
if err != nil {
	return nil, fmt.Errorf("yarax compiler: %w", err)
}

# loop through rule sources and add them to the compiler
for ... {
  ...
  # loop through rules and add them to the compiler
  if err := yxc.AddSource(string(bs), yarax.WithOrigin(path)); err != nil {
  	return fmt.Errorf("failed to parse %s: %v", path, err)
  }
}

# build rules
yrs := yxc.Build()

# create new scanner with built rules
scanner := yarax.NewScanner(rules)
fc, err := os.ReadFile(path)
if err != nil {
	return nil, err
}

# scan file contents with scanner
res, err := scanner.Scan(fc)
if err != nil {
	return nil, err
}
...
```

FWIW, I don't see any difference in behavior between `scanner.Scan` or `rules.Scan(fc)` since the latter is just creating a new scanner behind the scenes. Could this just be due to reading said larger files into memory and then scanning them or is there something else at play?

plusvic:
Each WASM module tries to reserve a large amount of virtual address space (4GB in 64-bits systems). This doesn't means that they actually needs this amount of RAM, it only means that it reserves address space in case the memory required by the module grows dynamically. As each YARA-X scanner needs one of these WASM modules, this implies that YARA-X ends up reserving a very large amount of virtual address space. This is problematic in some environments where the amount of virtual address space per process is limited (in Docker containers, for instance). 

This PR reduces the amount of virtual address space reserved by each WASM module, mainly by limiting the amount of memory of each WASM module to 16MB. This is enough to accommodate very large sets of rules, with millions of rules and patterns.

Closes #292.

### DIFF
--- a/lib/src/scanner/mod.rs
+++ b/lib/src/scanner/mod.rs
@@ -262,7 +262,7 @@ impl<'r> Scanner<'r> {
         // Create module's main memory.
         let main_memory = wasmtime::Memory::new(
             wasm_store.as_context_mut(),
-            MemoryType::new(mem_size, None),
+            MemoryType::new(mem_size, Some(mem_size)),
         )
         .unwrap();
 
--- a/lib/src/wasm/mod.rs
+++ b/lib/src/wasm/mod.rs
@@ -720,7 +720,8 @@ lazy_static! {
         config.native_unwind_info(false);
 
         config.cranelift_opt_level(wasmtime::OptLevel::SpeedAndSize);
-        config.memory_reservation(0x80000000); // 2GB
+        config.memory_reservation(0x10000000); // 256MB
+        config.memory_reservation_for_growth(0);
         config.guard_before_linear_memory(false);
         config.memory_may_move(false);
         config.epoch_interruption(true);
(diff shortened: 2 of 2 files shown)


egibs:
Thanks for looking into this!

PR #297, opened 2025-01-31
fix: Remove broken tests.

wxsBSD:
With the API changes to the linter we currently do not have a good way to expose this in the python API, so remove these failing tests for now.


wxsBSD:
I'm removing these for now, until we can come up with a way to expose the required metadata in a nicer way.

plusvic (approved):

PR #304, opened 2025-02-07
feat: Add tag linter.

wxsBSD:
Add a tag linter that can be configured to use either a list of allowed strings, or a regexp. If both are provided in the config file the allowed list is preferred by the "check" command as it is more explicit.

As part of this work I realized we have no way to return multiple linter findings in a single check call, which is necessary when checking tags, so I modified the internal API to return a `Option<Vec<LinterResult>` where no linter findings is represented by returning None.

I also discovered that the way the compiler works is that any linter errors are collected but we only return the first one when calling `add_source()` which is not ideal when you can have multiple linter errors collected up. To deal with this I added a `linter_errors: Vec<CompileError>` to the compiler struct so that they can be checked for separately from actual compiler errors.

I also had to modify the output handling of the `check` command so that multiple linter errors are handled correctly.

Lastly, a couple of typos and copy/paste mistakes in documentation were fixed.


wxsBSD:
I haven't added tests for this yet, but am happy to do so once you're happy with these changes.

I initially had it so that linter warnings were in their own collection, like I am doing for linter errors, but decided against that for now. It is conceptually nicer to have a list of compiler warnings and errors separate from a list of linter warnings and errors, so that they can be treated differently by different programs. I'm happy to move linter warnings into their own list if you like that idea more.

plusvic:
> As part of this work I realized we have no way to return multiple linter findings in a single check call, which is necessary when checking tags, so I modified the internal API to return a `Option<Vec<LinterResult>` where no linter findings is represented by returning None.
> 
> I also discovered that the way the compiler works is that any linter errors are collected but we only return the first one when calling `add_source()` which is not ideal when you can have multiple linter errors collected up. To deal with this I added a `linter_errors: Vec<CompileError>` to the compiler struct so that they can be checked for separately from actual compiler errors.

I wouldn't change the compiler's behavior with respect to errors. The current behaviour is that `Compiler::add_source` will return the first error it encounters while compiling the source code. A single call to `Compiler::add_source` ] *can* produce more than one error, and add them to the list that is obtained later via `Compiler::errors()`, but the compiler doesn't make any guarantees about finding all possible errors. When an error occurs, the compiler can abort immediately and refuse to keep compiling the current set of rules (in must cases that's what it does). So, in the case of the tags linter, if more than one tag fails to comply with the requirements, it's ok if the compiler only returns the first error. Also, I wouldn't make distinctions between compiler errors and linter errors, I prefer treating all errors in the same way, I think it's simpler and easier to understand that way. Telling a linter to produce errors means that it produces compiling errors as the compiler would do in all other cases. 

Warnings are a different story. It's true that the current API doesn't allow a linter to produce multiple warnings, and that's a limitation. But instead of returning `Option<Vec<LinterResult>>` from the linter, 
I would add another variant to `LinterResult`, we alredy have `Ok`, `Warn` and `Err`, we could also add `Warns(Vec<Warnings>)`, for cases in which the linter wants to return more than one warning. The linter can still using the `Warn` variant of returning a single warning.

wxsBSD:
Ok, I think I've addressed your initial round of comments. I'll add a couple of tests and this should be ready for review.

lib/src/compiler/errors.rs
```diff
@@ -739,6 +741,68 @@ pub struct InvalidRuleName {
...
+/// 1 | rule a : foo {
+///   |          ^^^ tag `foo` not in allowed list
+///   |
+///   = note: Allowed tags: test, bar
+/// ```
+#[derive(ErrorStruct, Clone, Debug, PartialEq, Eq)]
+#[associated_enum(CompileError)]
+#[error(
+    code = "E040",
+    title = "Tag not in allowed list"
```
plusvic, line 762: Error and warning titles always start with lowercase.

lib/src/compiler/errors.rs
```diff
@@ -739,6 +741,68 @@ pub struct InvalidRuleName {
...
+///   |
+/// 1 | rule a : foo {
+///   |          ^^^ tag `foo` does not match regex `bar`
+///   |
+/// ```
+#[derive(ErrorStruct, Clone, Debug, PartialEq, Eq)]
+#[associated_enum(CompileError)]
+#[error(
+    code = "E041",
+    title = "Tag does not match regex `{regex}`"
```
plusvic, line 793: Start title with lower case.

lib/src/compiler/linters.rs
```diff
@@ -114,6 +115,145 @@ impl LinterInternal for RuleName {
...
+///     r#"warning[unknown_tag]: Tag not in allowed list
+///  --> line:1:12
+///   |
+/// 1 | rule foo : test { strings: $foo = "foo" condition: $foo }
+///   |            ---- tag `test` not in allowed list
+///   |
+///   = note: Allowed tags: foo, bar"#);
+pub struct Tags {
+    allow_list: Vec<String>,
+    regex: String,
```
plusvic, line 144: I would make this an `Option<String>` and initialize it with `None`. That's more efficient that allocating an empty string during initialization, and also suggests that this string is optional, which is the case.

lib/src/compiler/linters.rs
```diff
@@ -114,6 +115,145 @@ impl LinterInternal for RuleName {
...
+            compiled_regex,
+            error: false,
+        };
+        Ok(tags)
+    }
+
+    /// Specifies whether the linter should produce an error instead of a
+    /// warning.
+    ///
+    /// By default, the linter raises warnings about rule names that don't match
```
plusvic, line 178: The comment must be updated:  "By default, the linter raises warnings about *tags*..."

lib/src/compiler/linters.rs
```diff
@@ -114,6 +115,145 @@ impl LinterInternal for RuleName {
...
+        if !self.allow_list.is_empty() {
+            for tag in tags.iter() {
+                if !self.allow_list.contains(&tag.name.to_string()) {
+                    if self.error {
+                        return LinterResult::Err(errors::UnknownTag::build(
+                            report_builder,
+                            tag.span().into(),
+                            tag.name.to_string(),
+                            Some(format!(
+                                "Allowed tags: {}",
```
plusvic, line 208: Start message with lowercase.

lib/src/compiler/linters.rs
```diff
@@ -114,6 +115,145 @@ impl LinterInternal for RuleName {
...
+                                self.allow_list.join(", ")
+                            )),
+                        ));
+                    } else {
+                        results.push(warnings::UnknownTag::build(
+                            report_builder,
+                            tag.span().into(),
+                            tag.name.to_string(),
+                            Some(format!(
+                                "Allowed tags: {}",
```
plusvic, line 218: Start message with lowercase.

lib/src/compiler/linters.rs
```diff
@@ -114,6 +115,145 @@ impl LinterInternal for RuleName {
...
+                        ));
+                    }
+                }
+            }
+        }
+
+        return if results.is_empty() {
+            LinterResult::Ok
+        } else {
+            return LinterResult::Warns(results);
```
plusvic, line 252: The `return` keyword can be removed here, you are already returning the result from th `if` statement.

plusvic (changes requested):

Issue #244, opened 2024-11-12
Ignore errors on rule compile


DigiAngel:
Topic.  I'd like to get a feature where we tell yara-x to ignore any rules with errors and continue on with the ones that do work.  Thank you.

DigiAngel:
Brilliant...thank you!

PR #307, opened 2025-02-18
Capture compilation errors to avoid affecting the entire process

Rabb1tQ:
#244


plusvic:
I'm not sure this is the feature described in #244. This ignores whole source files, including the rules that were compiled correctly but were contained in a file that failed to compile. Also this have additional implications, now the `yr compile` exits with error code 0 even if compilation errors were found, changing the current semantics.

PR #348, opened 2025-05-09
Add module invoke support to python bindings.

wxsBSD:
This commit adds the ability to dump module output from python bindings. The API is roughly:

yara_x.Module('pe').invoke(data)

I chose to use an API that requires you instantiate a new Module object because it will allow us to put other things as methods on it. I'm thinking we will want to support things like invoke a module on a path or file object. This sets us up to have a cleaner API in the future.

I also moved and renamed the SupportedModules enum out of the dump command so it is now in a better place to be shared by the bindings and dump command.  This allows us to raise a ValueError if you try to instantiate a new module with an unsupported dump module.

As part of this I also realized that when using "dump" and outputting YAML (the default) the keys would be different from outputting in JSON. For example, in YAML a PDB path would be "pdb_path" while in JSON it was "pdbPath". To better align these I made the JSON output use proto_field_true when serializing the protobuf to JSON.

### DIFF
--- a/lib/src/modules/mod.rs
+++ b/lib/src/modules/mod.rs
@@ -136,7 +136,25 @@ pub mod mods {
     # use yara_x;
     let pe_info = yara_x::mods::invoke::<yara_x::mods::PE>(&[]);
     ```
-     */
+    */
+
+    use clap::ValueEnum;
+    use strum_macros::{Display, EnumString};
+    #[derive(Debug, Clone, ValueEnum, Display, EnumString, PartialEq)]
+    #[strum(ascii_case_insensitive)]
+    /// Modules supported by the "dump" command and language bindings.
+    pub enum SupportedDumpModules {
+        /// LNK module for parsing lnk files.
+        Lnk,
+        /// Macho module for parsing macho files.
+        Macho,
+        /// Elf module for parsing elf files.
+        Elf,
+        /// PE module for parsing pe files.
+        Pe,
+        /// Dotnet module for parsing dotnet files.
+        Dotnet,
+    }
 
     /// Data structures defined by the `dotnet` module.
     ///
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4660,6 +4660,7 @@ dependencies = [
  "bitflags 2.9.0",
  "bitvec",
  "bstr",
+ "clap",
  "const-oid",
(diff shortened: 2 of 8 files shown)


wxsBSD:
If you are happy with this I can extend the API before merging this to allow for things like:

```
module = yara_x.Module('pe')
module.invoke_path('/some/path/to/exe')
module.invoke_file(open_file_object)
```

Or to combine it into `module.invoke()` with args to handle what you want to scan.

lib/src/modules/mod.rs
```diff
@@ -136,7 +136,25 @@ pub mod mods {
...
     ```
-     */
+    */
+
+    use clap::ValueEnum;
+    use strum_macros::{Display, EnumString};
+    #[derive(Debug, Clone, ValueEnum, Display, EnumString, PartialEq)]
+    #[strum(ascii_case_insensitive)]
+    /// Modules supported by the "dump" command and language bindings.
+    pub enum SupportedDumpModules {
```
plusvic, line 146: I'm not convinced about adding this enum to the public API. I understand that this derives from the need of using the enum both in the CLI and Python library, but from the API standpoint this enum is confusing because it won't be used in any of the API functions. Also, the "dump" command is a CLI thing, the API doesn't know about the dump.

The advantage of using this enum is that, once you add some item to it, you are forced to handle the new module in both the CLI and the Python library, the compiler will remind you to add the corresponding branch to the `match` statements. But this enum in the API feels unnatural to me. I would rather use the module name for deciding which module to invoke, even if this means that the Python library and the CLI are not forced to be in-sync with respect to the modules they support.

plusvic:
> If you are happy with this I can extend the API before merging this to allow for things like:
> 
> ```
> module = yara_x.Module('pe')
> module.invoke_path('/some/path/to/exe')
> module.invoke_file(open_file_object)
> ```
> 
> Or to combine it into `module.invoke()` with args to handle what you want to scan.

Let's keep the API simple as a starting point, we always have time to add more methods later, but removing them is more difficult .

lib/src/modules/mod.rs
wxsBSD: I've moved this back and created a new enum only visible to the python bindings.

plusvic (approved):

Issue #362, opened 2025-06-05
yara-x fails to compile with minimal features


demoray:
With the following features:

```toml
yara-x = { version = "1.0", default-features = false, features = ["constant-folding", "exact-atoms", "fast-regexp", "math-module"] }
```

consumers see the following error:

```
error[E0432]: unresolved import `nom`
 --> /home/USER/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/yara-x-1.0.0/src/types/func.rs:3:5
  |
3 | use nom::AsChar;
  |     ^^^ use of unresolved module or unlinked crate `nom`
  |
= help: if you wanted to use a crate named `nom`, use `cargo add nom` to add it to your `Cargo.toml`
```

plusvic:
Fixed via ae0f436.

PR #363, opened 2025-06-05
require nom due to it's use in lib/src/types/func.rs

demoray:


demoray:
This PR addresses #362

PR #368, opened 2025-06-10
feat: Better module error handling and error propagation

vojone:
I noticed that if error occurs in a module (e. g. caused by some syntax error in its (meta)data) it does not propagate the error to the user, but it just doesn't hit anything. In my opinion, it can be very misleading during the debugging process.

Therefore, I've got an idea to change the return type of module `MainFn` to `Result<Box<dyn MessageDyn>, std::string::String>`, which allow to return a string with description of an error which appeared inside a module.

For example if there is missing `,` in cuckoo JSON report we now get:
```
$ ./target/debug/yr scan -x cuckoo=cuckoo.json test.yar test.yar
error:  scanning "test.yar": cannot parse data for cuckoo: expected `,` or `}` at line 6 column 5
────────────────────────────────────────────────────────────────────────────────────────────────────────
 0 file(s) scanned in 0.3s. 0 file(s) matched.
```


plusvic:
I like this change, but I would be more strict about the errors returned by modules. Instead of returning `Result<Box<dyn MessageDyn>, std::string::String>>`  I would return some `Result<Box<dyn MessageDyn>, ModuleError>>`, where `ModuleError` would be a enum derived from `thiserror::Error`, like in:

https://github.com/VirusTotal/yara-x/blob/15a06bcc825dd2071cc2466cc12fff823c794caf/lib/src/compiler/errors.rs#L17-L38

This implies that new modules may need to add new variants to `ModuleError`, but this would be an internal API.

lib/src/modules/mod.rs
```diff
@@ -213,18 +215,24 @@ pub mod mods {
     /// # use yara_x;
     /// let elf_info = yara_x::mods::invoke::<yara_x::mods::ELF>(&[]);
     /// ```
-    pub fn invoke<T: protobuf::MessageFull>(data: &[u8]) -> Option<Box<T>> {
+    pub fn invoke<T: protobuf::MessageFull>(
+        data: &[u8],
+    ) -> Option<Result<Box<T>, String>> {
```
plusvic, line 220: This is a backward incompatible change in the API, and returning `Option<Result<...>` doesn't look like a good API. I think this API  can be kept as is.

vojone: Ok. I added conversion from `Result` to `Option`, so the signature remains the same to preserve the API.

vojone:
Thank you for the review. I agree with remarks, I tried to incorporate them.

PR #405, opened 2025-07-17
feat: Allow underscores in integer and float literals.

### LINKED ISSUE #14: Allow using underscores (_) as part of numeric literals

wxsBSD:
This commit allows underscores in both integer and float literals. You can now do things like:

```
uint32(0) == 0xaa_bb_cc_dd
uint32(0) == 65_535
```

Fixes #14.

### DIFF
--- a/parser/src/tokenizer/mod.rs
+++ b/parser/src/tokenizer/mod.rs
@@ -444,26 +444,24 @@ enum NormalToken<'src> {
     ]
     Ident(&'src [u8]),
 
-    // Float literals. Single underscores are allowed to separate any digits but
-    // neither part of the float may start with an underscore.
+    // Float literals. Underscores are allowed to separate digits.
     #[regex(
         r#"(?x)                         # allow comments in the regexp
-            ([0-9]+_?)+                 # one or more digits or underscores
+            ([0-9]+_*)+                 # one or more digits or underscores
             \.                          # a dot
-            ([0-9_]+_?)+                # one more digits or underscores
+            ([0-9_]+_*)+                # one more digits or underscores
         "#,
         |token| token.slice())
     ]
     FloatLit(&'src [u8]),
 
-    // Integer literals. Single underscores are allowed to separate any digits
-    // but the literal can not start with an underscore.
+    // Integer literals. Underscores are allowed to separate digits.
     #[regex(
         r#"(?x)
            (
-             0x([a-fA-F0-9]+_?)+ |       # hexadecimal number
-             0o([0-7]+_?)+       |       # octal number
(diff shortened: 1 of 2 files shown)


wxsBSD:
There is one thing that may be an issue with this. Using underscores in metadata results in them being stripped from the AST but not if they are used in a condition:

```
wxs@mbp yara-x % cat rules/test.yara
rule test {
  meta:
    a = 1_000
    b = 1_0.1_0
  condition:
    0xaa_bb == 0xaabb
    and 0.1_0 == 0.1
}
wxs@mbp yara-x % ./target/debug/yr debug ast rules/test.yara
 rule test
 ├─ meta
 │  ├─ a = 1000
 │  └─ b = 10.1
 └─ condition
    └─ and
       ├─ eq
       │  ├─ 0xaa_bb
       │  └─ 0xaabb
       └─ eq
          ├─ 0.1_0
          └─ 0.1


wxs@mbp yara-x %
```

It can be fixed but would require operating on the literal and not value when converting the CST to AST in cst2ast.rs (https://github.com/VirusTotal/yara-x/blob/main/parser/src/ast/cst2ast.rs#L609-L616) which requires expanding the MetaValue::Integer and MetaValue::Float to include the literal. I'm not sure how much of an actual bug this is though.

parser/src/tokenizer/mod.rs
```diff
@@ -447,9 +447,9 @@ enum NormalToken<'src> {
     // Float literals
     #[regex(
         r#"(?x)                         # allow comments in the regexp
-            [0-9]+                      # one or more digits
+            [0-9_]+                      # one or more digits
```
plusvic, line 450:
```suggestion
            [0-9_]+                      # one or more digits or underscores
```

plusvic:
We need a test case to make sure that `_0123` is tokenized as `Token::IDENT` and not as `Token::INTEGER_LIT`.

parser/src/tokenizer/mod.rs
```diff
@@ -447,9 +447,9 @@ enum NormalToken<'src> {
...
             \.                          # a dot
-            [0-9]+                      # one more digits
+            [0-9_]+                      # one more digits
```
plusvic, line 452:
```suggestion
            [0-9_]+                      # one more digits or underscores
```

plusvic (changes requested):

PR #413, opened 2025-07-22
fix: Add scanner class to python stub.

wxsBSD:
I think this was just overlooked when updating the python stub.


plusvic:
Oops, I didn't notice this PR before starting working on https://github.com/VirusTotal/yara-x/commit/d1ea0b0f70c210bd6d2916f73571bac6b876c814

PR #417, opened 2025-07-28
feat: implement macho export parsing from symbol table when DYLD_INFO and E…

latonis:
This allows for a more complete parsing of exports in macho files, especially those that do not have `LC_DYLD_INFO` or `LC_DYLD_EXPORTS_TRIE`. We then need to iterate the symbol table to find specific tags to identify the exports in the macho file.

Spoke with @pstirparo about the fallback behavior about how exports are surfaced in missing `LC_DYLD_INFO` or `LC_DYLD_EXPORTS_TRIE`.

This also results in a slight reorder of how symbols are parsed from the symbol/string tables. We now follow the index order from the symbol table `nlist` structures instead of parsing the string table directly.

This method is also seen in nm, otool, and dyld_info (links below for ref):
- https://github.com/apple-oss-distributions/dyld/blob/main/other-tools/dyld_info.cpp#L560-L617
- https://github.com/apple-oss-distributions/dyld/blob/main/mach_o/NListSymbolTable.cpp#L203-L220

Added a new test file for parsing these types of exports if they exist: 
- https://github.com/VirusTotal/yara-x/pull/417/files#diff-e3907dcfa118f6f7b2cae26530c8b2027f855188d9a408b3482dedfb9a3a63dd

### DIFF
--- a/lib/src/modules/macho/parser.rs
+++ b/lib/src/modules/macho/parser.rs
@@ -1108,7 +1108,8 @@ impl<'a> MachOFile<'a> {
         )
     }
 
-    /// Parser that parses the nlist structures that are defined by the offsets in LC_SYMTAB
+    /// Parser that parses the nlist structures that are defined by the offsets
+    /// in LC_SYMTAB.
     fn nlist(
         &self,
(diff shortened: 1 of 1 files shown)


pstirparo:
Thanks for being extremely responsive, that was a fast patch. LGTM.

PR #420, opened 2025-07-29
feat: implement transitive includes

PeterMatula:
The `include` implementation introduced to YARA-X is not the same as in YARA. The most notable difference is that it does not support transitive includes. I.e. if file `A` includes `B`, `B` cannot then include more files relative to its location (I'm not sure if absolute includes work in such a case). We need this transitive behavior in order to migrate to YARA-X in Gen Digital (Avast).

This PR attempts to add this functionality, related changes, improvements, and tests. However, I'm not 100% sure what is the best way to introduce this stuff, so this is kind of a draft. I fully expect you might have some notes or reservations. I kindly ask you to have a look and point me to an acceptable solution. I will try to accommodate the best I can. I believe the newly introduced tests demonstrate what I try to achieve.

### DIFF
--- a/lib/src/compiler/ir/tests/mod.rs
+++ b/lib/src/compiler/ir/tests/mod.rs
@@ -2,6 +2,7 @@ use std::mem::size_of;
 
 use crate::compiler::{Expr, IR};
 use crate::types::TypeValue;
+use crate::SourceCode;
 
 #[test]
 fn expr_size() {
@@ -91,7 +92,10 @@ fn ir() {
             path.with_extension("no-folding.ir")
         };
 
-        let source = fs::read_to_string(path).unwrap();
+        let source = fs::read_to_string(&path).unwrap();
+        let source_code = SourceCode::from(source.as_str()).with_origin(
+            &path.as_os_str().to_str().expect("Path should be valid UTF-8"),
+        );
 
         let output_file = mint.new_goldenfile(&output_path).unwrap();
         let mut compiler = Compiler::new();
@@ -100,7 +104,7 @@ fn ir() {
         compiler
             .hoisting(false)
             .set_ir_writer(w)
-            .add_source(source.as_str())
+            .add_source(source_code.clone())
             .unwrap();
 
         #[cfg(feature = "constant-folding")]
@@ -113,7 +117,7 @@ fn ir() {
             compiler
                 .hoisting(false)
                 .set_ir_writer(w)
-                .add_source(source.as_str())
+                .add_source(source_code.clone())
                 .unwrap();
 
             let hoisting_output = output_path.with_extension("hoisting.ir");
@@ -124,7 +128,7 @@ fn ir() {
             compiler
                 .hoisting(true)
                 .set_ir_writer(w)
-                .add_source(source.as_str())
+                .add_source(source_code.clone())
                 .unwrap();
         }
     });
--- a/lib/src/compiler/ir/tests/testdata/10-dependency.yar
+++ b/lib/src/compiler/ir/tests/testdata/10-dependency.yar
@@ -1,0 +1,4 @@
+rule test_10_dep {
+  condition:
+    true
+}
--- a/lib/src/compiler/ir/tests/testdata/10.cse.ir
+++ b/lib/src/compiler/ir/tests/testdata/10.cse.ir
@@ -1,0 +1,6 @@
+RULE test_10_dep
+  0: CONST boolean(true) -- parent: None 
+
+RULE test
+  0: SYMBOL Rule(RuleId(0)) -- parent: None 
+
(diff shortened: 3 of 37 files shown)


plusvic:
The current behavior for includes in YARA-X goes as follow:

* Before compilation you specify one or more include directories for the compiler to use.
* If no include directories are specified, the current directory is used by default.
* When looking for included files, the compiler will try to find them in any of those include directories, in the order they were specified.

If you have a file `foo.yar` that contains `include "bar.yar"`, the file `bar.yar` will be found as long as it is located in any of the include directories (or in the current directory if you didn't specify the include directories you want to use). If `bar.yar` contains another include like `include "baz.yar"`, the same logic applies and `baz.yar` will be found if its located in any of the include directories.

This differs from YARA, but it is simpler to implement and more aligned to how other compilers work, like C/C++ compilers. With this behavior you also have control over the directories that YARA-X will try to read files from. With YARA you can write rules that try to include arbitrary files from your file system, and YARA will read those files and try to load them as a rules file (sort of "escaping" from the directories where the rules are).

Can you provide more information about your current use case? If you can give the compiler the right list of include directories you can compile your rules, even if they contain includes that are relative to the file that contains them.

PeterMatula:
> Can you provide more information about your current use case?

We have thousands of YARA source files organized in complex directory structure with 100+ directories. Any rule author can decide to create whatever new structure as they see fit. The included files may include more sources, relative to their path, without any limitations. I.e. there is no immutable set of directories that can be determined. Migrating to a flat structure would not only be very strenuous, but it would be a downgrade to the current state, and frankly not acceptable by our rule authors and owners. Altough, I maybe see (the discussion below) what you mean.

> If you can give the compiler the right list of include directories you can compile your rules, even if they contain includes that are relative to the file that contains them.

I read this like 10 times, looked in to current code base, did some experiments, and I'm still not sure if this is true, or if I'm missing something. More precisely this part: 

> even if they contain includes that are relative to the file that contains them

[This](https://github.com/VirusTotal/yara-x/blob/c83ca69d31ce1f249ca072331633faf0a1289e9c/lib/src/compiler/mod.rs#L1204) function searches for files to include only from `.` or the specified include directories. There is not concept of `relative to the file that contains them`, is there? Everything is relative to the original include directories.

> more aligned to how other compilers work, like C/C++ compilers

Well, in C/C++ you can do `-I`, but the source files can also then include other files relative to them, no? Altough, here I probably got what you mean. I'm guessing you intend it to be used in [this style](https://github.com/VirusTotal/yara/blob/77f866f4aaef52e740665622d359027c5fe3f1a8/libyara/ahocorasick.c#L33-L38) as opposed to [this style](https://github.com/VirusTotal/yara/blob/77f866f4aaef52e740665622d359027c5fe3f1a8/sandbox/collect_matches.cc#L30-L33), as if the YARA rules were "installed", like C/C++ project would be? I guess it would be possible to use this in our code base, and I personally would even like it more as I prefer something like this:

```c
#include <project/whatever/include.h>
```

over something like this:

```c
#include "../../../../whatever/include.h"
```

which is what we currently have. I will bring this up with our rule owners, but given the amount of changes that would be required both in the current code base, and in all the associated tools and pipelines, and enforcing this style for the new rules going forward, I don't thing we will go for it at this time anyway :-(

We would either have to maintain this proposed change in our downstream fork, or find an acceptable variant of this that you would be willing to merge. In the end, the core of this is not that complicated, its [these few lines](https://github.com/VirusTotal/yara-x/pull/420/files#diff-0f7c520fb21eb875ca08339186f1d8220abead03fa1ccdbacf6d8ea2dde2c28aR1257-R1265). The rest are either tests, or some extra fail-safes -- some of which are (more) necessary because of the new transitive includes feature, some even the original implementation would benefit from. For example, the current upstream compiler gets into stack overflow on this (and as I play with it, even this PR does :-():

```yara
// file main.yar
include "main.yar"

rule test {
    condition: true
}
```

While this example is pretty synthetic and perhaps not very realistic, you can do the same with multiple files and `--include-dir` options. I would certainly be preferable for compiler to catch this and end nicely. It could also be beneficial in general (not just for our use case) for YARA-X to support these transitive/relative includes.

plusvic:
So, I guess you are probably in the worst possible case, with a ton of different directories from where files are included, and with a ton of include statements that are relative to the current file.

> If you can give the compiler the right list of include directories you can compile your rules, even if they contain includes that are relative to the file that contains them.

What I meant with this is that if you have a file `a.yar` that does `include "b.yar"`, it will work as long as you add the directory that contains `b.yar` to the list of include directories.

> Well, in C/C++ you can do -I, but the source files can also then include other files relative to them, no?

Yes, that's correct, it's true that in C/C++ you can do `#include "../../../../whatever/include.h"`, what I was referring to was the `#include <project/whatever/include.h>` style in combination with the `-I` command-line option.  This is the style that I would like to encourage to use. This way all include statements are relative to a small set of include directories, preferably a single one, which is the project root path.

What if I add a `yr fix imports` command that automatically fixes all the import statements and make them relative to some directory specific directory? Would you be able to apply such changes to your repository?

plusvic:
Already implemented in 89674f13d0092e1b715505fe497ce70573cdb0b1.

PR #421, opened 2025-07-29
feat: parse macho imports from symbol table when dyld_info or chainFixups are missing

latonis:
Similar to #417 but for imports.

Spoke with @pstirparo about the fallback behavior about how imports are surfaced in missing LC_DYLD_INFO or LC_DYLD_CHAINED_FIXUPS.

This method is also seen in nm, otool, and dyld_info (links below for ref):
- https://github.com/apple-oss-distributions/dyld/blob/main/other-tools/dyld_info.cpp#L372-L398
- https://github.com/apple-oss-distributions/dyld/blob/main/mach_o/NListSymbolTable.cpp#L321-L326


pstirparo:
Thanks for taking up this one too :)

PR #422, opened 2025-08-02
fix(macho): avoid duplicating imports if in weak_bind_off

latonis:
In some cases, if a symbol was defined in `LC_DYLD_INFO` or `LC_DYLD_INFO_ONLY`, an import could be defined twice via being defined in `bind_off` _and_ `weak_bind_off`, resulting in duplicative entries in the list of imports.

To address this, I added a `seen` imports HashSet to avoid adding duplicate entries to the table. This does not change import hash results as we were already deduplicating in the hash function.

For reference: https://github.com/apple/darwin-xnu/blob/main/EXTERNAL_HEADERS/mach-o/loader.h#L1318-L1349

```cpp
    /*
     * Some C++ programs require dyld to unique symbols so that all
     * images in the process use the same copy of some code/data.
     * This step is done after binding. The content of the weak_bind
     * info is an opcode stream like the bind_info.  But it is sorted
     * alphabetically by symbol name.  This enable dyld to walk 
     * all images with weak binding information in order and look
     * for collisions.  If there are no collisions, dyld does
     * no updating.  That means that some fixups are also encoded
     * in the bind_info.  For instance, all calls to "operator new"
     * are first bound to libstdc++.dylib using the information
     * in bind_info.  Then if some image overrides operator new
     * that is detected when the weak_bind information is processed
     * and the call to operator new is then rebound.
     */
    uint32_t   weak_bind_off;	/* file offset to weak binding info   */
    uint32_t   weak_bind_size;  /* size of weak binding info  */
```

Emphasis on 
```
This enable dyld to walk all images with weak binding information in order and look for collisions.
```

This brings us in line with nm, otool, etc on how we display imports as a list as well.


plusvic (approved):

pstirparo:
Great job @latonis, thanks for being super reactive with all those macho fixes.

Issue #114, opened 2024-05-21
Expanding support for global variable objects.


msuiche:
This is something which was lacking in the C version, according to the current [documentation yara-x](https://virustotal.github.io/yara-x/docs/writing_rules/external-global-variables/) only supports "global external variable" but not "external object" which would include arrays and structures for richer data enrichment.

This could be really great for modules that would want to keep the same name convention of VirusTotal live hunting to make rules interoperable for example variables like `vt.behaviour.command_executions` or `vt.behaviour.modules_loaded` which are only accessible as an array via the `for` loop keyword. But also variables under specific structures such as `vt.behaviour`.

More information about existing issues which were not addressable in the current C version of yara:
[Support for EXTERNAL OBJECT_TYPE_ARRAY and OBJECT_TYPE_STRUCTURE](https://github.com/VirusTotal/yara/issues/1960)
[Exporting yr_object_create() to enable custom structures?](https://github.com/Hugal31/yara-rust/issues/127)

plusvic:
This is something that I have in mind for future releases. I'm currently focused in releasing a version that has feature parity with `yara`, but more advanced features like this are in the plan.

msuiche:
Excellent! Amazing job as usual @plusvic !

nmbarel:
Seems like the rust api for Yara-x already supports global hashmap vars (actually for any type implementing try_into, unfortunately vectors do not implement it, so adding support for arrays will require much more, potentially breaking, code refactoring). I’ve added support for the python API as well in #433 
More work for the the rest of the APIs is needed, might try to tackle it myself after the PR is merged

msuiche:
@nmbarel just to confirm, does that mean you can have nested variable names with this update? like x.y.z.variable name?

nmbarel:
@msuiche 
Yep :) any dict should work, (example: global var dict={x: {y: 1}} should match with the rule condition: dict.x.y == 1 ) 

As I’ve said, I’m currently working on the rest of the apis and I think I might’ve been wrong about my list comment, looks like a valid serde json value can also be just a list so I’ll try getting on that after finishing the rest

msuiche:
Amazing! This is pretty cool, at the time I was looking for this I realized it could really unlock a lot of possibilities for third party even beyond security.

nmbarel:
[...]

PR #449, opened 2025-09-15
feat: Dict as global vars go c apis

nmbarel:
added dict support to the rest of the APIs


nmbarel:
Here is the rest of the APIs, as promised in #114, a quick question - @plusvic - I am not sure how the CLI works and how to add to it the support, if the C api should do it by its own or if not, where can I add that? Thanks in advance :)

nmbarel:
> Here is the rest of the APIs, as promised in #114, a quick question - @plusvic - I am not sure how the CLI works and how to add to it the support, if the C api should do it by its own or if not, where can I add that? Thanks in advance :)

nevermind, somehow completely missed the cli dir...
I'll get on that soon

PR #458, opened 2025-09-24
feat: implement `dex` module

delvinru:
Hi, I have implemented the `dex` module.

The current version is incompatible with the old version, because in that implementation there were not very convenient methods and names of exported variables.

What has been changed:
- Added new method `checksum` for computing adler-32 checksum;
- Added new method `signature` for computing sha-1 signature;
- Added new methods: `contains_strings`, `contains_methods`, `contains_classes` - honestly, I doubt their usefulness, but I decided to leave them, because binary search is used under the hood, since the strings are guaranteed to be sorted by the format itself;
- Methods removed: `has_method_string`, `has_method_and_class_string`, `has_method_regexp`, `has_method_and_class_regexp`, `has_class_string`, `has_class_regexp` (as far as I can see on github, no one used them anyway); And in practice, quite useless calls;
- Changed access to strings, methods, types, protos and classes. I've made them more informative and user-friendly, unlike the previous implementation;
- Added documentation, because on https://yara.readthedocs.io / there is no information about `dex` module;

Fuzzing testing was started locally, all errors were fixed, but I do not exclude that something else will be revealed during a long run.

Since the most difficult thing in programming is to come up with a name for variables, if there are suggestions about names in proto, then I'm ready for discussion.


plusvic (changes requested):

delvinru:
@plusvic Hi, there are a lot of clippy errors in the pipeline that are not related to my code (or any my changes)

It's better to fix them in the master, and then merge them into this branch, rather than the other way around.

delvinru:
@plusvic Hi, does this MR look okay, or do I need to add something else to merge it?

plusvic:
> @plusvic Hi, does this MR look okay, or do I need to add something else to merge it?

I'm running the fuzzer a little bit before merging it.

delvinru:
@plusvic hi, any updates?

lib/src/modules/dex/parser.rs
```diff
@@ -0,0 +1,789 @@
...
+        data: &'a [u8],
+        header: &DexHeader,
+    ) -> IResult<&'a [u8], Vec<Rc<String>>> {
+        // DEX file doesn't contain strings.
+        // It's a strange case, but it needs to be checked.
+        if header.string_ids_off == 0 {
+            return Ok((remainder, Vec::new()));
+        }
+
+        let (rem, string_offsets) =
```
plusvic: We need more sanity checks here. The `string_ids_size` field is read directly from the file and can contain an arbitrarily large value, which will result in an arbitrarily large `Vec` that can exhaust memory. 

This file causes an OOM because of this.
[oom-c45a50821a11ab59fdc0a268e72eb22d394ff29a.zip](https://github.com/user-attachments/files/23603789/oom-c45a50821a11ab59fdc0a268e72eb22d394ff29a.zip)

lib/src/modules/dex/parser.rs
```diff
@@ -0,0 +1,789 @@
...
+        string_items: &[Rc<String>],
+    ) -> IResult<&'a [u8], Vec<Rc<String>>> {
+        // DEX file doesn't contain types.
+        // It's a strange case, but it needs to be checked.
+        if header.type_ids_off == 0 {
+            return Ok((remainder, Vec::new()));
+        }
+
+        let (rem, type_indexes) =
+            count(le_u32::<&[u8], Error>, header.type_ids_size as usize)
```
plusvic: This has the same issue mentioned above.

lib/src/modules/dex/parser.rs
plusvic: This is an example using an iterator. It doesn't creates a temporary vector containing the string offsets.

```rust
    fn parse_string_ids<'a>(
        remainder: &'a [u8],
        data: &'a [u8],
        header: &DexHeader,
    ) -> IResult<&'a [u8], Vec<Rc<String>>> {
        // DEX file doesn't contain strings.
        // It's a strange case, but it needs to be checked.
        if header.string_ids_off == 0 {
            return Ok((remainder, Vec::new()));
        }

        let mut it = iterator(remainder, le_u32::<&[u8], Error>);

        let string_ids = it
            .by_ref()
            .take(header.string_ids_size as usize)
            .filter_map(|offset| Self::parse_string_from_offset(data, offset))
            .map(Rc::new)
            .collect();

        let (remainder, _) = it.finish()?;

        Ok((remainder, string_ids))
    }
```

plusvic:
I think this is ready to merge. My only concern is the names of some fields like `proto_ids`, `string_ids`, `method_ids` and so on. My understanding is that these names are taken from the documentation that describes the DEX format, but I find identifiers like `protos`, `strings` and `methods` more straightforward and easy to read. In most cases I like the idea of following the same nomenclature used by the official documentation for the file format, as most people are already familiar with those names, but in this case I find the names very misleading. This is particularly true for the `string_ids` field, why `strings_ids`, when it is actually a list containing the strings themselves?

delvinru:
I agree, it was just an attempt to maintain some compatibility with the previous implementation.

In that case, it is better to make more understandable names right away. I'll fix them.

plusvic:
@delvinru check my latest changes and tell me what you think.

delvinru:
@plusvic It looks fine, I agree with the changes. I like the alternative with `.len()`, it looks much cleaner.

As mentioned in the commit message, if we need something in the future, we can just add it.

PR #478, opened 2025-10-17
feat: Show globals

nmbarel:
added a feature to show all global variables currently loaded into the compiler to help debugging issues, for Rust and the C API.

I am unsure about the current implementation, as i cannot really handle errors. having looked around a bit i thought the best approach would be to do something like how you handle the last error, however here i have a Vector to return, and i do not really know how to handle that through C...
I have added a small skeleton to the lib file, @plusvic I would love to know how you think this should be handled best, and how to handle a vector of strings passed from rust to C.

Thanks in advance :)


plusvic:
Can you explain what's your use case for this API?

nmbarel:
The endpoint will make debugging things a lot easier, as you can see in the symbols file I currently filter by field but we can widen it to check on any symbol in real time. 

Specifically, I would like to have access to the currently loaded envs as a part of my data pipeline, I do many switches between envs, and load/reload them per file I scan, so I would like to have a single source of truth to the current state of the compiler and its loaded components, this will also help me in testing said pipeline and rules therein to check for inconsistencies

plusvic:
I think we need to find a different way to accomplish this. The proposed API is simply dumping all the information in a string, and it includes internal details that are not guaranteed to remain unchanged in the future. One option is exporting all this information in JSON format, including only the names and values, but not internal details like `index`, `is_root`, etc.

nmbarel:
Yeah sure, this was part of the question I was asking in the original comment… I will
update it to a json with only the relevant data, but how do you suggest I should approach this? Kind of as implemented here? (As just another function) or more like the last_error implementation (where you update it constantly and allow access through a special function that returns a vector of strings)
Would love your input on this :)

nmbarel:
Hello @plusvic, I think i got it way closer to a finished up version :).
Would love to get your input on this, especially on the [lib/src/types/mod.rs] file, as I don't think it is the best way to do it but I am not really sure where else to take it...
(Also, if you have a better way of communicating that is not through git threads, We can move this conversation there :)

plusvic:
> Hello @plusvic, I think i got it way closer to a finished up version :). Would love to get your input on this, especially on the [lib/src/types/mod.rs] file, as I don't think it is the best way to do it but I am not really sure where else to take it... (Also, if you have a better way of communicating that is not through git threads, We can move this conversation there :)

I'm not quite convinced about the changes in `lib/src/types/mod.rs` neither. I don't have a more elegant solution now, but I guess it should be something about serializing a `TypeValue` into a JSON in a more idiomatic way.

nmbarel:
@plusvic, made the switch to serde_json as requested, I left Func as Null as i do not think there is really an idiomatic way to show them as json (also you cannot create a global variable of them :), let me know if any other changes are needed!

nmbarel:
Hey @plusvic, pinging this :)

PR #482, opened 2025-10-22
fix: Check address of functions for 0x0 in PE

vojone:
Added condition of parsing PE exports. Exports should not be parsed if `AddressOfFunctions` is 0x0, because:
1. Time/memory is wasted for nonse exports
2. Bad number of exports is provided (e. g. max number of PE exports)

This fix causes skip of the parsing of exports if `AddressOfFunctions` is 0x0 and therefore `pe.number_of_exports` is 0 in this case.


plusvic:
Do you have some file with `AddressOfFunctions` set to zero that we can add to the test cases?. It must be a small PE file, ideallistically a few kilobytes.

PR #491, opened 2025-11-04
fix: allow deprecated cargo_bin

wxsBSD:
81a4578aa03bd8980385d18fe355f05114c9cb36 updated assert_cmd to 2.1.1 and assert_cmd::Command::cargo_bin was deprecated[1] in 2.1.0 so now rust-analyzer is complaining about the usage. However, trying to move to the recommended assert_cmd::cargo::cargo_bin_name!() doesn't work because that uses an environment variable that is set at build time, and contains the path to the binary being built. However, cli tests need to run the "yr" binary which doesn't work with the recommended macro.

See https://docs.rs/assert_cmd/2.1.1/src/assert_cmd/macros.rs.html#85-92 for details, but the TLDR is that cargo_bin_cmd!() expects to find the path to the binary to run at an environment variable named CARGO_BIN_EXE_yr and that is not set. There may be some trickery to set it during the build but I haven't figured it out yet, nor do I know how to make this work in a platform independent way.

For now I'm just marking them deprecated.

[1]: https://docs.rs/assert_cmd/2.1.1/assert_cmd/cmd/struct.Command.html#method.cargo_bin

### DIFF
--- a/cli/src/tests/check.rs
+++ b/cli/src/tests/check.rs
@@ -26,6 +26,7 @@ fn metadata() {
         )
         .unwrap();
 
+    #[allow(deprecated)]
     Command::cargo_bin("yr")
         .unwrap()
         .arg("--config")
@@ -72,6 +73,7 @@ warning[text_as_hex]: hex pattern could be written as text literal
         )
         .unwrap();
 
+    #[allow(deprecated)]
     Command::cargo_bin("yr")
         .unwrap()
         .arg("--config")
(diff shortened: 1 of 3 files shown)


wxsBSD:
I'm not a big fan of doing this but I couldn't see a nice way to make this work otherwise. Please feel free to point me in a better direction if you know of one.

plusvic:
Fixed in #492 by setting the `CARGO_BIN_EXE_yr` environment variable. Running the CLI tests is not very straightforward anyways, requiring the environment variable to be set beforehand is acceptable. Now the CI includes these tests too.

Issue #500, opened 2025-11-20
Panic with a function call?


wxsBSD:
I haven't had time to debug this but wanted to get it out there as I was kind of surprised by the behavior. I was experimenting with weird ASTs due to my work in the deps branch and discovered it is possible to get a panic with a rule that I believe is syntactically valid if a bit a weird. ;)

This rule compiles and works correctly:

```
import "math"

rule a {
  condition:
    // This is a bit weird but allowed by the language...
    math.(entropy(0, 10)) > 3
}

```

But this rule causes a panic:

```
import "math"
import "pe"

rule a {
  condition:
    // If the previous rule is allowed by the language then this should be allowed too, right?
    math.(entropy(pe.sections[1].raw_data_offset, pe.sections[1].raw_data_size)) > 3
}
```

The panic I get is:

```
wxs@mbp yara-x % RUST_BACKTRACE=1 ./target/debug/yr scan rules/test2.yara /bin/ls

thread '<unnamed>' (25370463) panicked at lib/src/wasm/mod.rs:1135:17:
expecting field with index 1 in Struct {
    fields: {},
    is_root: false,
    protobuf_type_name: Some(
        "math.Math",
    ),
}
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/std/src/panicking.rs:698:5
   1: core::panicking::panic_fmt
             at /rustc/ed61e7d7e242494fb7057f2657300d9e77bb4fcb/library/core/src/panicking.rs:75:14
   2: yara_x::wasm::lookup_field::{{closure}}
             at ./lib/src/wasm/mod.rs:1135:17
   3: core::option::Option<T>::unwrap_or_else
             at /Users/wxs/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/option.rs:1061:21
   4: yara_x::wasm::lookup_field
             at ./lib/src/wasm/mod.rs:1134:14
   5: yara_x::wasm::lookup_object
             at ./lib/src/wasm/mod.rs:1180:22
   6: core::ops::function::Fn::call
             at /Users/wxs/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ops/function.rs:79:5
[...]

PR #502, opened 2025-11-26
fix: make the parser stricter.

plusvic:
The parser was too permissive and accepted expressions like `pe."foo"` and `pe.-1` as valid. This permissiveness could result in issues like the one described in #500. This PR introduces some changes in the parser to make it reject such invalid expressions.

### DIFF
--- a/parser/src/ast/cst2ast.rs
+++ b/parser/src/ast/cst2ast.rs
@@ -1456,7 +1456,14 @@ where
                     exprs.push(self.primary_expr()?);
                 }
 
-                let mut final_exprs: Vec<Expr> = exprs
+                // Consecutive dot-separated expressions can be coalesced
+                // into a single one based on the operands types. For
+                // example `a.b` (identifier . identifier), is merged into
+                // one FieldAccess expression. Similarly, if the left side
+                // is already a FieldAccess, and the right is an identifier,
+                // the identifier is simply added to the operand list of
+                // the FieldAccess. There are more cases, all covered below.
+                let mut coalesced: Vec<Expr> = exprs
                     .into_iter()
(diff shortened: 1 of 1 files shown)


Srylax:
@plusvic how is someone expected to access json struct data loaded from global env? Example:
we load json data that may look the following:
```json
{
  "config": {
    "fileSystems": {
      "/tmp": {
        "options": ["nodev"]
      },
      "/": {}
    }
  }
}

```
`config.fileSystems."/tmp"` is loaded into the env but cannot be accessed.

plusvic:
> @plusvic how is someone expected to access json struct data loaded from global env? Example: we load json data that may look the following:
> 
> ```json
> {
>   "config": {
>     "fileSystems": {
>       "/tmp": {
>         "options": ["nodev"]
>       },
>       "/": {}
>     }
>   }
> }
> ```
> 
> `config.fileSystems."/tmp"` is loaded into the env but cannot be accessed.

This is an interesting case, but it goes beyond YARA's goal. YARA allows defining structures using JSON, but never intended to allow access to fields that don't look like identifiers. This should probably be an error.

Srylax:
> @plusvic This should probably be an error.

Well it actually is. It does not let me compile the rule which is very annoying because now I have to preprocess the data to remove these characters

plusvic:
> > @plusvic This should probably be an error.
> 
> Well it actually is. It does not let me compile the rule which is very annoying because now I have to preprocess the data to remove these characters

I was thinking more about an error when calling `set_global`. If some key name is not a valid identifier it could fail, in the same way that it fails if you pass an array where not all the items have the same type. In JSON you can have `[1,true,"foo"]`, but that's not a valid array in YARA, and  `set_global` will fail in such case.

PR #508, opened 2025-12-04
feat: initial Language Server Protocol implementation

prosperritty:
## Overview

The Language Server Implementation provides a public function `yara_x_ls::serve` for launching, which is used in `cli` crate. This function allows to specify various communication channels. In the CLI tool these channels are `stdin` and `stdout`. The Language Server is launched using the CLI tool as follows:

```
yr ls
````

Most LSP features providing certain operations related to symbols were implemented: [Goto Definition](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition), [Find References](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_references), [Document Highlights](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentHighlight), [Hover](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover), [Document Symbols](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentSymbol), [Completion](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion), [Rename](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_rename). Unfortunately, these features can correctly process the identifiers of rules and patterns only. Support for symbols from `with` and `for` statements is currently being tested.

Other LSP features that were implemented are: [Selection Range](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_selectionRange) (based on the `CST` structure), [Semantic Tokens](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens), [PublishDiagnostics](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_publishDiagnostics) and [Pull Diagnostics](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_pullDiagnostics). Diagnostics currently only provides syntactic error analysis obtained from `AST` and minimal semantic checks for patterns in rules (usage, existence).

In most cases, it is necessary (or preferable) to take into account the punctuation of the language, which is why `CST` is most often used in the implementation for specific analysis instead of `AST`.

All implemented features currently operate within a single file. This means that the language server, for example, cannot process symbols included from another file. Nevertheless, it works consistently with multiple files (opened in some development tool) and can distinguish them in separate requests from the client.

Current implementation does not spawn threads.

The reason why the Language Server `Cargo.toml` has such `default-features` is to allow compilation into `wasm32-unknown-unknown`. With the `default-features` disabled, the Language Server can be compiled using `wasm-pack`.

## Implementation details

The Language Server is implemented using [`async-lsp`](https://github.com/oxalica/async-lsp) crate. This crate was inspired by another popular crate for creating Language Servers [`tower-lsp`](https://github.com/ebkalderon/tower-lsp). But `async-lsp` fixes some of its issues and bugs, such as immutable server state and incorrect notifications handling(see [this section](https://github.com/oxalica/async-lsp?tab=readme-ov-file#tower-lsp)). For reference, a fairly popular [Language Server for `nix`](https://github.com/oxalica/nil) was implemented using this crate.

### Brief project structure description

- `src/server.ts` 
  - Contains a structure for the server state that implements the `async_lsp::LanguageServer` trait, allowing `async_lsp::router::Router` to map client LSP requests and notifications to individual handlers.
- `src/utils/position.rs`
  - Contains conversion between LSP [`Position`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position)/[`Range`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range) types and YARA-X parser `Span` type.
- `src/utils/cst_traversal.rs`
  - Contains implementation of reusable `CST` traversals for searching for a declaration or occurrences of a specific symbol (rule or pattern identifier).
- `src/features/*.rs`
  - Contains implementation of various LSP features.

## Testing

Each test case is designed to verify the correct behavior of a specific implemented LSP feature under specified parameters. The tests check the expected behavior of the server (and the implementation of a specific module) during client-server communication, rather than calling a specific function from the `src/features` module. Therefore implemented testing can be considered as unit testing and partially E2E testing. Each test case has the following workflow:

1. Language Server (`yara-x-ls`) and mock Language Client are created and launched in separate threads
2. Communication is initialized([`Initialize`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize) and [`Initialized`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialized) messages are sent to the server)
3. Language Client sends a [`textDocument/didOpen`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didOpen) notification which contains the file content for a specific test case
4. Language Client sends a request of a specific type ([`textDocument/definition`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition), [`textDocument/hover`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover), etc.) and then checks that the server has returned the expected response
5. Language Server receives a request to shutdown, after which it is checked that all threads have terminated without errors.

All test cases are defined in `src/tests/testdata/` as follows:
- `tested_feature_name/source.yar` is YARA source code for a specific test case.
- `tested_feature_name/params_and_expected.json` contains a definition of the parameters with which the client will send a request(`"params"` field) and the expected response from the server(`"expected"` field).

If you have a suggestion on how to test that the Language Server can be used within `yr` binary (`cli` test case), please let me know.
Thanks.

### DIFF
--- a/ls/src/features/completion.rs
+++ b/ls/src/features/completion.rs
@@ -8,18 +8,14 @@ use crate::features::completion_const::{
     CONDITION_SUGGESTIONS, PATTERN_MOD, RULE_KW_BLKS, SRC_SUGGESTIONS,
 };
 
-use crate::utils::{cst_traversal::rule_from_pos, position::to_abs};
+use crate::utils::cst_traversal::rule_from_span;
 
 /// Provides completion suggestions based on the cursor position and the
 /// block it is in.
-pub fn completion(
-    cst: CST,
-    text: &str,
-    pos: Position,
-) -> Option<CompletionResponse> {
-    let pos_span = to_abs(pos, text);
-
-    let completion_cursor = cst.root().token_at_offset(pos_span as usize);
+pub fn completion(cst: CST, pos: Position) -> Option<CompletionResponse> {
+    let completion_cursor = cst
+        .root()
+        .token_at_position((pos.line as usize, pos.character as usize));
     let completion_token: Token<Immutable>;
 
     // Extract token before cursor
@@ -79,7 +75,7 @@ pub fn completion(
     let completion_vec = match nearets_section_kind {
         SyntaxKind::SOURCE_FILE => Some(source_file_suggestions()),
         SyntaxKind::CONDITION_BLK | SyntaxKind::BOOLEAN_EXPR => {
-            condition_suggestions(cst, completion_token, &pos_span)
+            condition_suggestions(cst, completion_token)
         }
         SyntaxKind::PATTERN_DEF => pattern_suggestions(completion_token),
         SyntaxKind::RULE_DECL => Some(rule_suggestions()),
@@ -93,7 +89,6 @@ pub fn completion(
 fn condition_suggestions(
     cst: CST,
     completion_prev_token: Token<Immutable>,
-    pos_span: &u32,
 ) -> Option<Vec<CompletionItem>> {
     let mut completion_vec: Vec<CompletionItem> = Vec::new();
 
@@ -138,7 +133,7 @@ fn condition_suggestions(
         | SyntaxKind::PATTERN_COUNT
         | SyntaxKind::PATTERN_OFFSET
         | SyntaxKind::PATTERN_LENGTH => {
-            let rule = rule_from_pos(&cst, pos_span)?;
+            let rule = rule_from_span(&cst, &completion_prev_token.span())?;
 
             let patterns = rule
                 .children()
--- a/ls/src/features/document_highlight.rs
+++ b/ls/src/features/document_highlight.rs
@@ -6,9 +6,9 @@ use yara_x_parser::cst::{SyntaxKind, CST};
 use crate::utils::{
     cst_traversal::{
         pattern_from_condition, pattern_from_strings, rule_from_condition,
-        rule_from_ident, rule_from_pos,
+        rule_from_ident, rule_from_span,
     },
-    position::{to_abs, to_range},
+    position::{node_to_range, token_to_range},
 };
 
 // The document highlight request is sent from the client to the server to
@@ -17,12 +17,11 @@ use crate::utils::{
 // ranges of all occurrences of that symbol in the source code.
 pub fn document_highlight(
     cst: CST,
-    text: &str,
     pos: Position,
 ) -> Option<Vec<DocumentHighlight>> {
-    let pos_span = to_abs(pos, text);
-
-    let highlight_cursor = cst.root().token_at_offset(pos_span as usize)?;
+    let highlight_cursor = cst
+        .root()
+        .token_at_position((pos.line as usize, pos.character as usize))?;
 
     match highlight_cursor.kind() {
         //Find highlight of pattern within the same rule
@@ -32,7 +31,7 @@ pub fn document_highlight(
         | SyntaxKind::PATTERN_LENGTH => {
             let mut highlight_vec: Vec<DocumentHighlight> = Vec::new();
 
-            let rule = rule_from_pos(&cst, &pos_span)?;
+            let rule = rule_from_span(&cst, &highlight_cursor.span())?;
 
             let declaration =
                 pattern_from_strings(&rule, highlight_cursor.text());
@@ -41,18 +40,22 @@ pub fn document_highlight(
                 pattern_from_condition(&rule, highlight_cursor.text());
 
             if let Some(declaration) = declaration {
-                highlight_vec.push(DocumentHighlight {
-                    range: to_range(declaration.span(), text),
-                    kind: Some(DocumentHighlightKind::WRITE),
-                });
+                if let Some(range) = node_to_range(&declaration) {
+                    highlight_vec.push(DocumentHighlight {
+                        range,
+                        kind: Some(DocumentHighlightKind::WRITE),
+                    });
+                }
             }
 
             if let Some(occurrences) = occurrences {
                 for occurrence in occurrences {
-                    highlight_vec.push(DocumentHighlight {
-                        range: to_range(occurrence.span(), text),
-                        kind: Some(DocumentHighlightKind::READ),
-                    });
+                    if let Some(range) = token_to_range(&occurrence) {
+                        highlight_vec.push(DocumentHighlight {
+                            range,
+                            kind: Some(DocumentHighlightKind::READ),
+                        });
+                    }
                 }
             }
 
@@ -65,10 +68,12 @@ pub fn document_highlight(
             let rule = rule_from_ident(&cst, highlight_cursor.text());
 
             if let Some(rule) = rule {
-                highlight_vec.push(DocumentHighlight {
-                    range: to_range(rule.span(), text),
-                    kind: Some(DocumentHighlightKind::WRITE),
-                });
+                if let Some(range) = node_to_range(&rule) {
+                    highlight_vec.push(DocumentHighlight {
+                        range,
+                        kind: Some(DocumentHighlightKind::WRITE),
+                    });
+                }
             }
 
             let occurrences =
@@ -76,10 +81,12 @@ pub fn document_highlight(
 
             if let Some(occurrences) = occurrences {
                 for occurrence in occurrences {
-                    highlight_vec.push(DocumentHighlight {
-                        range: to_range(occurrence.span(), text),
-                        kind: Some(DocumentHighlightKind::READ),
-                    });
+                    if let Some(range) = token_to_range(&occurrence) {
+                        highlight_vec.push(DocumentHighlight {
+                            range,
+                            kind: Some(DocumentHighlightKind::READ),
+                        });
+                    }
                 }
             }
 
--- a/ls/src/features/goto.rs
+++ b/ls/src/features/goto.rs
@@ -1,16 +1,16 @@
 use crate::utils::cst_traversal::{
-    pattern_from_strings, rule_from_ident, rule_from_pos,
+    pattern_from_strings, rule_from_ident, rule_from_span,
 };
-use crate::utils::position::{to_abs, to_range};
+use crate::utils::position::node_to_range;
 use async_lsp::lsp_types::{Position, Range};
 use yara_x_parser::cst::{SyntaxKind, CST};
 
 /// Return the range of the definition of a symbol at the specified
 /// position in the text if exists.
-pub fn go_to_definition(cst: CST, text: &str, pos: Position) -> Option<Range> {
-    let pos_span = to_abs(pos, text);
-
-    let goto_cursor = cst.root().token_at_offset(pos_span as usize)?;
+pub fn go_to_definition(cst: CST, pos: Position) -> Option<Range> {
+    let goto_cursor = cst
+        .root()
+        .token_at_position((pos.line as usize, pos.character as usize))?;
 
     #[allow(irrefutable_let_patterns)]
     match goto_cursor.kind() {
@@ -20,16 +20,16 @@ pub fn go_to_definition(cst: CST, text: &str, pos: Position) -> Option<Range> {
         | SyntaxKind::PATTERN_COUNT
         | SyntaxKind::PATTERN_OFFSET
         | SyntaxKind::PATTERN_LENGTH => {
-            let rule = rule_from_pos(&cst, &pos_span)?;
+            let rule = rule_from_span(&cst, &goto_cursor.span())?;
 
             let pattern = pattern_from_strings(&rule, goto_cursor.text())?;
 
-            Some(to_range(pattern.span(), text))
+            node_to_range(&pattern)
         }
         //Rule identifiers
         SyntaxKind::IDENT => {
             let rule = rule_from_ident(&cst, goto_cursor.text())?;
-            Some(to_range(rule.span(), text))
+            node_to_range(&rule)
         }
         _ => None,
     }
--- a/ls/src/features/hover.rs
+++ b/ls/src/features/hover.rs
@@ -5,9 +5,8 @@ use yara_x_parser::cst::{
     Immutable, Mutable, Node, NodeOrToken, SyntaxKind, CST,
 };
 
-use crate::utils::{
-    cst_traversal::{pattern_from_strings, rule_from_ident, rule_from_pos},
-    position::to_abs,
+use crate::utils::cst_traversal::{
+    pattern_from_strings, rule_from_ident, rule_from_span,
 };
 
 /// Builder for hover markdown representation of a rule.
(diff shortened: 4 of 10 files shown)

### CODE REVIEW

**Comment by plusvic** [comment]
Is it possible to use the language server from VSCode?

**Comment by prosperritty** [comment]
> Is it possible to use the language server from VSCode?

Yes, but I haven't added the extension setup for VS Code. I was not sure if it would be appropriate to add it here.
Would you like to have extension setup within `ls` crate?

**Comment by plusvic** [comment]
> > Is it possible to use the language server from VSCode?
> 
> Yes, but I haven't added the extension setup for VS Code. I was not sure if it would be appropriate to add it here. Would you like to have extension setup within `ls` crate?

It would be helpful. I would like to have at least some basic documentation for setting up an environment that I can use for testing the language server in real life.

**Comment by prosperritty** [comment]
> > > Is it possible to use the language server from VSCode?
> > 
> > 
> > Yes, but I haven't added the extension setup for VS Code. I was not sure if it would be appropriate to add it here. Would you like to have extension setup within `ls` crate?
> 
> It would be helpful. I would like to have at least some basic documentation for setting up an environment that I can use for testing the language server in real life.

I've added simple VS Code extension setup for launching Extensions Development Host that includes language server. Basic documentation on how to launch this can be found in `yara-x/ls/editors/code/README.md`. 

After last step in `README.md` just open some `.yar`/`.yara` file and it should work :)

**Comment by plusvic** [comment]
I've reviewed the code and noticed that there are a lot functions that receive both the AST and the original text because it is needed in order to translate spans to (line,col) pairs required by the LSP. 

The CST already allows to obtain a node by its absolute offset within the text, and for each node it provides its span within the text, but it doesn't provide its (line,col) position nor obtain the node at a given (line,col) position . However I believe this could be done, and with that the code for the language server would be a lot more straightforward and less prone to errors. All those (line,col) to span conversions (and viceversa) would be unnecessary, as the CST would provide that information.

I will give it a try too see if I can come up with a solution for that.

**Comment by prosperritty** [comment]
> I've reviewed the code and noticed that there are a lot functions that receive both the AST and the original text because it is needed in order to translate spans to (line,col) pairs required by the LSP.
> 
> The CST already allows to obtain a node by its absolute offset within the text, and for each node it provides its span within the text, but it doesn't provide its (line,col) position nor obtain the node at a given (line,col) position . However I believe this could be done, and with that the code for the language server would be a lot more straightforward and less prone to errors. All those (line,col) to span conversions (and viceversa) would be unnecessary, as the CST would provide that information.
> 
> I will give it a try too see if I can come up with a solution for that.

That would be great! Thank you.

**Comment by plusvic** [comment]
> > I've reviewed the code and noticed that there are a lot functions that receive both the AST and the original text because it is needed in order to translate spans to (line,col) pairs required by the LSP.
> > The CST already allows to obtain a node by its absolute offset within the text, and for each node it provides its span within the text, but it doesn't provide its (line,col) position nor obtain the node at a given (line,col) position . However I believe this could be done, and with that the code for the language server would be a lot more straightforward and less prone to errors. All those (line,col) to span conversions (and viceversa) would be unnecessary, as the CST would provide that information.
> > I will give it a try too see if I can come up with a solution for that.
> 
> That would be great! Thank you.

I've implemented a `token_at_line_col` method (c7f9f55), which is similar to `token_at_offset`, but it receives a line, column pair instead of an offset. I've also implemented a `line_col` method (a251d12) that returns the line and column where a token or node starts. 

This allows to simplify the language server, as you don't need to track the original source code, now the CST is enough. It also has the advantage that even if you modify the CST (for instance by inserting new code), you still can get line and column numbers that takes into account the modifications that have been made.

**Comment by prosperritty** [comment]
> I've implemented a `token_at_line_col` method ([c7f9f55](https://github.com/VirusTotal/yara-x/commit/c7f9f557a919c3b2a8f4504f37d6b3be4b14329b)), which is similar to `token_at_offset`, but it receives a line, column pair instead of an offset. I've also implemented a `line_col` method ([a251d12](https://github.com/VirusTotal/yara-x/commit/a251d12bbced9a25c6d000d8dd74fc6244a52b07)) that returns the line and column where a token or node starts.
> 
> This allows to simplify the language server, as you don't need to track the original source code, now the CST is enough. It also has the advantage that even if you modify the CST (for instance by inserting new code), you still can get line and column numbers that takes into account the modifications that have been made.

Great, I'll try to integrate the new API into my implementation as soon as possible. I'll also add support for non-unix `stdin`/`stdout` streams, due to the absence of which test fails.

**Comment by prosperritty** [comment]
I think I've found a bug in the new API. It does not take into account new lines within multiline comments. Let's consider the following code and calling `token_at_line_col` function with (line, col) of `<C>` :

```yara
/*
Multi-line <C>comment
Another line
*/
rule simple {
  condition: // single-line comment
    true
}
```

This function will only count new lines in the source code and ignore those found in comments, since new lines in comments are part of the token text itself and are not independent tokens:

```
Current token it the loop: COMMENT@0..37 "/*\nMulti-line comment ..."
Current token it the loop: NEWLINE@37..38 "\n"
Current token it the loop: RULE_KW@38..42 "rule"
Current token it the loop: WHITESPACE@42..43 " "
Current token it the loop: IDENT@43..49 "simple"
Current token it the loop: WHITESPACE@49..50 " "
Returned token: WHITESPACE@49..50 " "
```

Can I try to fix it in this PR?

plusvic:
> I think I've found a bug in the new API. It does not take into account new lines within multiline comments. Let's consider the following code and calling `token_at_line_col` function with (line, col) of `<C>` :
> 
> ```yara
> /*
> Multi-line <C>comment
> Another line
> */
> rule simple {
>   condition: // single-line comment
>     true
> }
> ```
> 
> This function will only count new lines in the source code and ignore those found in comments, since new lines in comments are part of the token text itself and are not independent tokens:
> 
> ```
> Current token it the loop: COMMENT@0..37 "/*\nMulti-line comment ..."
> Current token it the loop: NEWLINE@37..38 "\n"
> Current token it the loop: RULE_KW@38..42 "rule"
> Current token it the loop: WHITESPACE@42..43 " "
> Current token it the loop: IDENT@43..49 "simple"
> Current token it the loop: WHITESPACE@49..50 " "
> Returned token: WHITESPACE@49..50 " "
> ```
> 
> Can I try to fix it in this PR?

Oh, good catch! Don't worry about it, I'll fix it. You can focus in the language server itself.

plusvic:
I also noticed that a position in LSP has zero-based line and column numbers, so I'll do the same in the API. And I will also make the API more flexible and allow column numbers based in byte offsets and Unicode code points. These are important details when dealing with unicode characters.

prosperritty:
> I also noticed that a position in LSP has zero-based line and column numbers, so I'll do the same in the API. And I will also make the API more flexible and allow column numbers based in byte offsets and Unicode code points. These are important details when dealing with unicode characters.

That would be more convenient for me. I thought that the non zero-based API was more intuitive, so I didn't ask for it. Thank you :)

prosperritty:
@plusvic , can I add refactoring using the new `CST` API now?

I also noticed that you removed the `default-features` because the build was failing. Could we make some features from `asynс-lsp` and `tokio` crate optional somehow so it could be compiled to WASM? I could try to find another possible way without `default-features` and suggest it to you.

plusvic:
> @plusvic , can I add refactoring using the new `CST` API now?
> 
> I also noticed that you removed the `default-features` because the build was failing. Could we make some features from `asynс-lsp` and `tokio` crate optional somehow so it could be compiled to WASM? I could try to find another possible way without `default-features` and suggest it to you.

Sure, I just took the easiest way, but I would like to be able to compile the language server to WASM. Which are the options in `async-lsp` and `tokio` that prevents the compilation to WASM? 

Also, I moved the language server to its own binary, removing the library in the process. But I guess that in order to run it in WASM we need to have the language the library server as a library, right?

prosperritty:
> Sure, I just took the easiest way, but I would like to be able to compile the language server to WASM. Which are the options in `async-lsp` and `tokio` that prevents the compilation to WASM?

If I remember correctly, in order to compile the language server to WASM we have to disable:
- `tokio`, `client-monitor`, `stdio` features from `async-lsp` crate
-  entire `tokio` (and `tokio-util` as well) crate, since tokio runtime is not supported for WASM

I also noticed that I'm not using tower service from `client-monitor` feature in my current implementation, I just forgot to remove it. Therefore, we can remove this feature completely for now.

> Also, I moved the language server to its own binary, removing the library in the process. But I guess that in order to run it in WASM we need to have the language the library server as a library, right?

Yes, it has to be a library. Initially, I had a `lib.rs` containing the same `serve` function and `main.rs` that calls this function providing `stdio` channels as arguments. It worked, but I'm not sure it is the best way to do it. Nevertheless, if you consider this approach appropriate, I can do the same thing here.

I would also like to note that compilation of `ls` crate to WASM does not make the language server itself possible to run by calling `serve` function from JS. There has to be another function (meaning with the `#[wasm_bindgen]` attribute ) for that, which accepts JS objects as arguments. For example, I used something similar to what is shown in [`tower-lsp-web-demo`](https://github.com/silvanshade/tower-lsp-web-demo/blob/main/crates/browser/src/lib.rs) to run the language server using WASM with my `serve` function.

Also, since the language server is a separate binary now, I should change setup for VS Code extension and documentation on how to launch Extensions Development Host. I'll do it in the next commits :)

plusvic:
While investigating how the LSP client and server agree on a position encoding I found what looks like a roadblock. The protocol describes three kinds of position encoding: utf-8, utf-16, and utf-32 (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position). The client and server must agree on the position encoding they will use, so that they can interpret the locations in the same way.

From the server side, we could use either utf-8 or utf-32, but not utf-16. However, the client only seems to support utf-16 encoding, which is the default. If I add `position_encoding: Some(PositionEncodingKind::UTF32)`, to the server capabilities, the client starts to fail with this error:

```
Unsupported position encoding (utf-32) received from server YARA-X LSP
```

I can't find a way to indicate in the client that it should accept utf-32 position encoding. 

If we don't solve this issue, everything will be fine as long as you don't use unicode characters that are encoded as two bytes or more. But once you start using such characters, the client will interpret the position sent by the server incorrectly. For instance, with the current version of the server and client we get this:

image

This happens because the server is currently using utf-8 encoding (column numbers are simply byte offsets), but the client expects utf-16 (columns are the number of utf-16 code units).

plusvic:
> While investigating how the LSP client and server agree on a position encoding I found what looks like a roadblock. The protocol describes three kinds of position encoding: utf-8, utf-16, and utf-32 ([microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification#position](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position)). The client and server must agree on the position encoding they will use, so that they can interpret the locations in the same way.
> 
> From the server side, we could use either utf-8 or utf-32, but not utf-16. However, the client only seems to support utf-16 encoding, which is the default. If I add `position_encoding: Some(PositionEncodingKind::UTF32)`, to the server capabilities, the client starts to fail with this error:
> 
> ```
> Unsupported position encoding (utf-32) received from server YARA-X LSP
> ```
> 
> I can't find a way to indicate in the client that it should accept utf-32 position encoding.
> 
> If we don't solve this issue, everything will be fine as long as you don't use unicode characters that are encoded as two bytes or more. But once you start using such characters, the client will interpret the position sent by the server incorrectly. For instance, with the current version of the server and client we get this:
> 
> image
> This happens because the server is currently using utf-8 encoding (column numbers are simply byte offsets), but the client expects utf-16 (columns are the number of utf-16 code units).

It looks the only viable option is supporting the utf-16 position encoding in the server, which in turns means that we better implement it in the CST.

prosperritty:
> I can't find a way to indicate in the client that it should accept utf-32 position encoding.

I've also tried to find a way to set [`PositionEncodingKind`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#positionEncodingKind) for VS Code, but it seems impossible. The only place where changes to VS Code language client configuration can be done is here (in `extension.ts`):

```ts
let clientOptions: LanguageClientOptions = {
    documentSelector: [{ scheme: "file", language: "yara" }],
  };
```

`LanguageClientOptions` allows only to specify encoding for communication(channels), but not for position offsets. Another way to set this could be `settings.json`, but I couldn't find an option for that there as well. 

> It looks the only viable option is supporting the utf-16 position encoding in the server, which in turns means that we better implement it in the CST.

I'm using `AST` for `Diagnostics` feature, and the problem with position offset may appear in this feature as well (see screenshot below). I haven't studied how YARA-X parser handles different encodings, so I wanted to ask if adding support for utf-16 in `CST` will make it possible to use utf-16 in `AST` as well? I think `AST` is more suitable for `Diagnostics` feature, but I can rewrite it using `CST`, which unfortunately may affect the quality of the analysis.

image

Getting line and column information directly from the AST will be imposible. In that case the only option storing the original source code and converting spans into line and column.

plusvic:
There are a lot of places with code like `self.documents.get(...).cloned();`. Source code can be very large, and those clone operations are expensive. They are not really required, see 61f42a6 and 67ab4a5.

prosperritty:
I rewrote most of the code where it was possible to use new `CST` position API.

I'll try to get rid of old approach in `Semantic Tokens` and `Diagnostics` where I use `CSTStream` and `AST` later. Since current implementation provides only so-called [**full**](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens_fullRequest) `Semantic Tokens` feature, I can easily count lines manually in loop. But I'm not sure what if there is some elegant way to rework `Diagnostics` feature (I hesitate we can count lines in `AST` as it is done in `CST`), but I'll try to see what I can do.

Also there is a `selection_range_comment` test that is failing, because `token_at_position` does not return multiline comment token in some cases (when cursor position within multiline comment is not the first line of the comment). I guess we can ignore this test for now?

plusvic:
I've merged this PR into the `language-server`. Let's continue the development in that branch, once it gets mature I will merge it into `main`.

plusvic:
@prosperritty I'll work on refactoring the `semantic_tokens_full` method, so that uses the new positions API, instead of spans. Once this is finished the renderization issues should be solved.

prosperritty:
> @prosperritty I'll work on refactoring the `semantic_tokens_full` method, so that uses the new positions API, instead of spans. Once this is finished the renderization issues should be solved.

Ok, but I'd like to remind you about VS Code constraints. VS Code does not support [`multilineTokenSupport`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokensClientCapabilities), and therefore semantic tokens for multiline comments must be separated into different tokens for each line. In my current implementation LS counting byte offsets when processing multiline comments within Semantic Tokens feature, so the renderization issue has to be solved there seperately. Thanks.

PR #515, opened 2025-12-18
feat: C api console log

wxsBSD:
Add support for setting the `console.log()` to the C API.

Special thanks to Victor for explaining closures and moving ownership with them properly.


MFMokbel:
Thank you, @plusvic and @wxsBSD for the quick implementation. Appreciated.

PR #530, opened 2026-01-20
feat(ls): improve access to module symbols

prosperritty:
This PR improves `Code Completion` feature and also implements `Signature Help`.  Now the language server suggests everything related to modules: their identifiers, functions, constants, nested structure identifiers, and regular fields. 

All the necessary information for this is stored statically in hash tables. These hash tables are generated in the build script. This script parses all files related to the modules, including Rust source code and protobuf files. 

The main motivation for adopting this approach is to access the entire hierarchy of modules using a single interface (in this case, hash tables), and not just their fields. Second, but no less important, is the goal of maintaining consistency with the documentation of individual functions or structures. In other words, store documentation of all module structures and functions that will be shown in pop up windows for `Hover`, `Code Completion` or `Signature Help` directly in the source code where they are implemented or declared. For functions, the description can be saved in the Rust documentation comments (`///`), and current implementation already extracts those documentation comments and then show them in the pop up window. For fields, it is possible to store documentation in directly protobuf as options (for example within `FieldOptions`), but I'm not sure it is the best approach.

I understand that this approach is relatively complex, and I believe that something from `lib` crate in `build.rs` can be reused for the language server to extract information about field symbols and reduce the source code of `build.rs`. But I'm not sure that it is possible to keep the documentation for them within main crate.

At the moment, the tests are failing due to non-deterministic sequence in the responses. If it is possible to keep the approach with hash tables, I can figure out how to avoid the test failure.

### DIFF
--- a/ls/build.rs
+++ b/ls/build.rs
@@ -253,7 +253,7 @@ fn __get_HASHMAP{hasmap_name}() -> &'static HashMap<&'static str, Vec<ItemModule
             item_module_value.push_str("None,");
         }
 
-        // Completiom item kind for Code Completion feature
+        // Completion item kind for Code Completion feature
         item_module_value.push_str(&format!(
             "kind: {},",
             self.kind.to_completion_item_kind()
@@ -454,7 +454,7 @@ fn find_message_by_full_name<'a>(
     }
 }
 
-// Processes all input protos for a specific module
+// Processes all input protobuf files for a specific module.
 fn process_proto_module_definition(
     yara_modules: &mut YaraModules,
     protos: &[FileDescriptorProto],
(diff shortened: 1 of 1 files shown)


plusvic:
I like the idea, but the current implementation has two main issues:

* Duplicates complexity that already exists in the main `yara-x` crate for enumerating functions exported by modules.
* `ls/build.rs` needs access to the source files in the `yara-x` crate. This works because we have all crates in the same repository, but violates the premise that a Rust crate is an independent unit that can be compiled on its own.

When I implemented field completion in https://github.com/VirusTotal/yara-x/commit/3f3110defe733a945b7d4e366dd083b3ef87ed03, what I did was exposing the names of the structures and fields in each module via a new API in the `yara-x` crate. My idea was extending this API in the future to include functions as well, and I think that this approach is better because it can reuse the mechanism that already exist in `yara-x` for enumerating functions, without having to duplicate all that complexity in the `yara-x-ls` crate.

plusvic:
Support for function in autocompletion has been partially added in 0c44b3f289b41a128dfec622a5596bf0a5a65295. I say partially because function documentation is shown, but improvements in that regard should follow the current approach of exposing the relevant information through the `mods::reflect` API.

PR #564, opened 2026-02-16
feat(ls): improvements regarding included files

prosperritty:
This PR improves the following LSP language features, that from now can recognize rule identifiers from included files:

- Find References
- Hover
- Rename
- Code Completion

All logic for that is implemented in and centered around `DocumentStorage` structure. With this approach, the server can first check whether the user has opened the file and not read it from the file system in this case. This also separates the logic of working with the file system from the implementation of the language features themselves. It is more convenient because conditional compilation will be managed in one place (so that it will be possible to compile to Wasm in the future).

For `Find References` and `Rename` it tries to find all occurrences of the rule identifier within (single) opened workspace. For `Hover` and `Go to definition` it tries to find the rule definition in all includes (transitive includes). For `Code Completion` it now additionally collects rule identifiers from the included files and suggests them in the condition block.

I must admit, that current implementation is not optimized at all. It most often traverses entire workspace to resolve necessary includes and even do double reads from FS. My idea was to start from something and now focus on how to improve the performance. At the moment, I am planning to try implementing a cache to solve this problem. But first of all, I would like to hear your opinion or plans so I won't waste time on implementing something that you don't want to see, as I see that you have already started to implement this. Thanks :)

### DIFF
--- a/ls/src/tests/mod.rs
+++ b/ls/src/tests/mod.rs
@@ -164,6 +164,8 @@ where
         let mut response_json = serde_json::to_value(actual_response).unwrap();
 
         replace_in_json(&mut response_json, test_dir.as_str(), "${test_dir}");
+        sort_json(&mut response_json);
+
         serde_json::to_writer_pretty(response_file, &response_json).unwrap();
         server_socket
     })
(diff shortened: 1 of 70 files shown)


plusvic:
`rename` now returns a `HashMap` instead of a `Vec`, the issue with that is that `HashMap` items don't have a stable order, and test cases fail due to the unpredictable order.

PR #571, opened 2026-02-23
feat(compiler): provide source line in the diagnostics report

metthal:
This will fill in the missing information to reconstruct a nice error / warnings for consumers of serialized diagnostics (for example in C API)

I thought long and hard on this and I think this is the best approach if I'm not missing anything. Basically, we'd like to use C API of YARA-X and extract compilation diagnostics from it. The problem is that C API exposes the serialized report where you have all the information about the issue + the `annotate-snippets` text, where the source line is embedded. However, if you'd like to build your own rich info diagnostic printer, you don't have the source line available.

Yes, you can cache the source code on the caller side but you probably don't want to do that if you have multiple files because YARA-X is also caching the source codes so doing it on 2 sides of the fence seems kind of redundant. This allows caller to not cache the source code, let YARA-X do the caching and extraction of the source lines and provide only the necessary parts.

If you think another solution is more appropriate, let me know. Thanks

### DIFF
--- a/lib/src/compiler/report.rs
+++ b/lib/src/compiler/report.rs
@@ -148,6 +148,22 @@ impl Report {
                     ),
                 };
 
+            // Extract the source line to include into diagnostics
+            let start = span.start();
+            let line_start = match code[..start].rfind('\n') {
+                Some(i) => i + 1,
+                None => 0,
+            };
+            let line_end = match code[start..].find('\n') {
+                Some(i) => start + i,
+                None => code.len(),
+            };
+
+            // Strip trailing CR from possible CRLF
+            let source_line = code[line_start..line_end]
+                .trim_end_matches('\r')
+                .to_string();
+
             Label {
                 level: level_as_text(level),
                 code_origin,
@@ -155,6 +171,7 @@ impl Report {
                 column,
                 span,
                 text,
+                source_line,
             }
         })
     }
@@ -353,6 +370,7 @@ pub struct Label<'a> {
     column: usize,
     span: Span,
     text: &'a str,
+    source_line: String,
 }
 
 impl Label<'_> {
--- a/lib/src/compiler/tests/mod.rs
+++ b/lib/src/compiler/tests/mod.rs
@@ -1136,7 +1136,8 @@ fn errors_serialization() {
                 "line": 1,
                 "column": 23,
                 "span": { "start": 22, "end": 25 },
-                "text": "this identifier has not been declared"
+                "text": "this identifier has not been declared",
+                "source_line": "rule test {condition: foo}"
             }
         ],
         "footers": [],


plusvic:
The issue with this approach is that errors may refer to code that spans multiple lines. The `span` field on each label is not guaranteed to be contained within a single line. So, I'm not quite convinced about this.

A better solution is including the whole text covered by the label's span, but that could be a lot of text, so I'm not convinced with this neither.

metthal:
Maybe I was just testing in on rather small (or wrong) examples but it was usually single line, so I just worked with assumption is always single line. There should be the whole span available though, so finding all the affected lines shouldn't be an issue, but I understand the worry about too much text.

However, there is already full text of `annotated-snippets` diagnostics already baked into the diagnostics JSON, so a possibility of too much text is already there. So what about this?

What if we completely omit the `annotated-snippets` text from the returned diagnostics JSON and instead replace it with just the source text? I'm honestly not sure whether it's not more useful than having the "rendered" error in there. But maybe someone is using it and this would break API.

plusvic:
> Maybe I was just testing in on rather small (or wrong) examples but it was usually single line, so I just worked with assumption is always single line. There should be the whole span available though, so finding all the affected lines shouldn't be an issue, but I understand the worry about too much text.

This is an example where the error covers multiple lines:

```
rule test_1 {
  condition:
     true
}

/*/
rule test_2
{
  strings:
    $a = "foo"
    $b = "bar"
    $c = "baz"
    $d = "qux"
  condition:
    all of them
}
```

The error message looks like:

```
error[E001]: syntax error
  --> test.yar:6:1
   |
 6 | / /*/
 7 | | rule test_2
 8 | | {
 9 | |   strings:
...  |
15 | |     all of them
16 | | }
   | |_^ unclosed comment
```

The error span goes from the comment's beginning to the end of the source code, because the comment is not closed. However, notice that the snippet omits lines 10 to 14, because they are not really relevant. It only shows some lines at the beginning, and some at the end of the span. No matter how many strings you put in the `strings` section, they won't appear in the error message, as they are not relevant.

> 
> However, there is already full text of `annotated-snippets` diagnostics already baked into the diagnostics JSON, so a possibility of too much text is already there. So what about this?
> 

Yes, the snippet is baked in the JSON, and that's a lot of text. But as it was shown above, `annotate-snippets` is smart enough to omit portions of the text that is not relevant.

> What if we completely omit the `annotated-snippets` text from the returned diagnostics JSON and instead replace it with just the source text? I'm honestly not sure whether it's not more useful than having the "rendered" error in there. But maybe someone is using it and this would break API.

I would say quite the opposite, the "rendered" error not only shows portions of the source code, it also shows puts the corresponding annotations in the source code. In many cases the user of the API wants to print the error into a text-mode console, and in such cases the annotated snippet is what you want. Producing these "rendered" snippets is not trivial, so the API is doing the heavy-lifting for you. 

Can you provide more details about your use case?

metthal:
My use-case was mainly colors which weren't present in the text but I somehow completely ignored existence of `YRX_COLORIZE_ERRORS` which I haven't noticed when I was looking at it for the first time. This probably resolves my issue - so we no longer have to reconstruct the error on our end if want to display it to the user if the colored way. We can directly use the colors of YARA-X with `YRX_COLORIZE_ERROS`. I guess this one solved itself :)

PR #579, opened 2026-03-03
fix: use dunce to canonicalize path in sli scan cmd

PeterMatula:
After migration, our users reported strange path prefixes with YARA-X that were not used with YARA. Looks like a known "problem" that [dunce](https://crates.io/crates/dunce) solves. There are not many canonicalizations going on in YARA-X, but I still applied this only to the one-and-only use in CLI, which is the most user-facing and I didn't want to touch compiler etc.

```
"matches": [
    {
      "rule": "hunting_interpreter_wscript",
      "file": "\\\\?\\C:\\temp\\samples\\CBDE39ACCBB9420E0F65BEDB225C89EF7909D1FEEFFDF467AA0EC8F435906E7B",
      ...
    },
    ...
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1209,6 +1209,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "dunce"
+version = "1.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


plusvic:
This was mentioned in #212. The decision back then was maintaining the UNC paths, but the `dunce` solutions seems like a good tradeoff.

PR #583, opened 2026-03-04
feat: added support for WASM builds using js-sys

niallnsec:
I have an (admittedly strange) need to run YARA in the browser as part of a rule management/threat hunting UI. The older C version of YARA could be compiled quite nicely to WASM using emscripten which worked well.
I wanted to upgrade to yara-x to take advantage of new features, but ran into some issues compiling yara-x to WASM. It would have been possible to use emscripten with the C API but that would mean having several levels of unnecessary abstractions and the associated performance issues. The only real blocker for a functional JS build was, ironically I thought, the dependency on wasmtime. Since a JS build will be running in a WASM environment already I figured the compiled rules could be executed directly in the same runtime context instead of using the wasmtime engine.

To do that I put together a replacement for the wasmtime crate to be used in the JS builds only that provides the same external interface while using the native WASM runtime environment internally.

I fully understand if you decide not to merge this PR if it is out of scope of what you want to support, but as its a lightweight change and I know others are also using the WASM builds of the C version of Yara (CyberChef as an example) I thought I would contribute back in the hopes it will be useful to others. I have an NPM package built using this modified fork which I will also publish seperately.

### DIFF
--- a/lib/src/wasm/runtime.rs
+++ b/lib/src/wasm/runtime.rs
@@ -9,7 +9,7 @@ mod browser_runtime {
     use std::pin::Pin;
     use std::sync::Arc;
 
-    use anyhow::{anyhow, Result};
+    use anyhow::{Result, anyhow};
     use js_sys::{
         Array, BigInt, Function, Object, Reflect, Uint8Array, WebAssembly,
     };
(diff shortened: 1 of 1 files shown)


plusvic:
Great contribution! Supporting WASM compilation was in my plans since long ago. My plan is merging this into a separate branch and continue developing this feature there, including a new crate that provides binding for JavaScript using  `wasm-bindgen`.

PR #591, opened 2026-03-14
feat(ls): add support for relative includes in diagnostics

prosperritty:
By default, the language server's `cwd` is the path to the open folder in the editor, which is also obtained in the compiler when diagnostics are requested. Therefore, the language server always reports an `include file not found` error when relative paths are used in `include` statement. Here, I've added a setting (`"YARA.relativeIncludePath"`) that, when enabled, will add the directory containing the file being processed in the compiler's include directories. In that case, users who use relative paths may receive correct diagnostics.

### DIFF
--- a/ls/src/configuration.rs
+++ b/ls/src/configuration.rs
@@ -13,8 +13,6 @@ pub struct Config {
     pub rule_name_validation: Option<String>,
     #[serde(default)]
     pub cache_workspace: bool,
-    #[serde(default)]
-    pub relative_include_path: bool,
 }
 
 /// This structure represents settings for the YARA-X formatter.
--- a/ls/src/features/diagnostics.rs
+++ b/ls/src/features/diagnostics.rs
@@ -65,11 +65,9 @@ pub fn compiler_diagnostics(
 
     let mut compiler = Compiler::new();
 
-    // Add path to the folder where the document is located when support
-    // for relative include is enabled in the configutaion
-    #[cfg(not(target_family = "wasm"))]
-    if config.relative_include_path
-        && let Ok(file_path) = document.uri.to_file_path()
+    // The folder where the document is located is added as an include
+    // directory in order to support relative include paths.
+    if let Ok(file_path) = document.uri.to_file_path()
         && let Some(folder_path) = file_path.parent()
     {
         compiler.add_include_dir(folder_path);
--- a/ls/src/tests/testdata/diagnostics7.yar
+++ b/ls/src/tests/testdata/diagnostics7.yar
@@ -1,4 +1,4 @@
-include "src/tests/testdata/diagnostics6.yar"
+include "diagnostics6.yar"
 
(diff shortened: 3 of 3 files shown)


prosperritty:
@plusvic There must be a `cfg` gate for `wasm` family target, otherwise it cannot be compiled into `wasm32-unknown-unknown`:

```
error[E0599]: no method named `to_file_path` found for struct `Url` in the current scope                                                                                                   
  --> ls/src/features/diagnostics.rs:70:41
   |
70 |     if let Ok(file_path) = document.uri.to_file_path()
   |                                         ^^^^^^^^^^^^ method not found in `Url`
```

And also, if we do not want to have it as a settings, we have to remove the same thing from `package.json` in VS Code extension.

And by the way, it looks like the workflow I've added for WASM LS build was not triggering all the time. I think we should fix that :(

plusvic:
Include paths are now always relative to the file that contains the include statement, which is the default behavior for the CLI tool. We could add some configuration option that allow specifying additional include directories, in case some user is working with a rule repository were all includes are relative to the base of that repository, but I guess this case is probably uncommon.

prosperritty:
> Include paths are now always relative to the file that contains the include statement, which is the default behavior for the CLI tool. We could add some configuration option that allow specifying additional include directories, in case some user is working with a rule repository were all includes are relative to the base of that repository, but I guess this case is probably uncommon.

I was also thinking about this, but I think it should not be set through the language server configuration. Since the user will have to change the settings manually every time they open, say, a different rule repository. 

However, we could discuss the idea of standardizing a specific manifest file in YARA-X language server. I know that YARA-X already has a [configuration file](https://virustotal.github.io/yara-x/docs/cli/config-file/), and perhaps we could integrate it somehow.

plusvic:
> @plusvic There must be a `cfg` gate for `wasm` family target, otherwise it cannot be compiled into `wasm32-unknown-unknown`:
> 
> ```
> error[E0599]: no method named `to_file_path` found for struct `Url` in the current scope                                                                                                   
>   --> ls/src/features/diagnostics.rs:70:41
>    |
> 70 |     if let Ok(file_path) = document.uri.to_file_path()
>    |                                         ^^^^^^^^^^^^ method not found in `Url`
> ```
> 
> And also, if we do not want to have it as a settings, we have to remove the same thing from `package.json` in VS Code extension.
> 
> And by the way, it looks like the workflow I've added for WASM LS build was not triggering all the time. I think we should fix that :(

Oh, I see. Fixed in 30aba26fc2b13c0dbfa37024e8c5d5c054c38306.

plusvic:
> I was also thinking about this, but I think it should not be set through the language server configuration. Since the user will have to change the settings manually every time they open, say, a different rule repository.

That's true. Let's keep it like this for now. I think that most people that use the include statement are using paths that are relative the the current file, not a common base directory.

PR #631, opened 2026-04-21
perf(wasm): avoid lowercase allocations in ASCII string ops

king-tero:
## Summary

  Optimize case-insensitive string operations for ASCII inputs.

  This avoids allocating lowercase copies for both operands in `icontains`,
  `istartswith`, `iendswith`, and `iequals` when both runtime strings are ASCII.
  For non-ASCII inputs the previous behavior is preserved.

  ## What changed

  In `lib/src/wasm/string.rs`:

  - added an ASCII fast path for:
    - `icontains`
    - `istartswith`
    - `iendswith`
    - `iequals`
  - kept the existing lowercase-based fallback for non-ASCII strings

  Also added a couple of non-ASCII regression checks in the generic string tests.

  ## Why

  The previous implementation always did this for case-insensitive string ops:

  - lowercase the left-hand side
  - lowercase the right-hand side
  - perform the comparison

  That is simple and correct, but expensive for the common ASCII case because it
  allocates and copies both strings on every call.

  This change keeps the same behavior while avoiding those temporary allocations
  for ASCII inputs.

  ## Results

  Benchmarked on a runtime ASCII string (~8.3 KB) with repeated operations:

  | Operation | Before | After | Speedup |
  |---|---:|---:|---:|
  | `icontains` | 14.758 ms | 0.283 ms | 52.1x |
  | `istartswith` | 14.762 ms | 0.275 ms | 53.7x |
  | `iendswith` | 14.705 ms | 0.277 ms | 53.1x |
  | `iequals` | 29.286 ms | 8.452 ms | 3.5x |

  These are targeted microbenchmarks for the string operations themselves, which
  is where the change applies.

  ## Validation

  Ran:

  - `cargo +1.91.0 fmt --all --check`
  - `cargo +1.91.0 check -p yara-x`
  - `cargo +1.91.0 test -p yara-x --lib string_operations -- --nocapture`
  - `cargo +1.91.0 test -p yara-x --lib test_proto2_module -- --nocapture`

  ## Notes

  Non-ASCII behavior is unchanged: those cases still go through the existing
  lowercase-based path.

### DIFF
--- a/lib/src/tests/mod.rs
+++ b/lib/src/tests/mod.rs
@@ -276,6 +276,8 @@ fn string_operations() {
 
     condition_true!(r#""foo" icontains "FOO""#);
     condition_true!(r#""foo" icontains "OO""#);
+    condition_true!(r#""CAFÉ" icontains "fé""#);
+    condition_true!(r#""mañana" istartswith "MAÑ""#);
     condition_true!(r#""foo" istartswith "Fo""#);
     condition_true!(r#""foo" iendswith "OO""#);
 
--- a/lib/src/wasm/string.rs
+++ b/lib/src/wasm/string.rs
@@ -242,9 +242,19 @@ impl RuntimeString {
         case_insensitive: bool,
     ) -> bool {
         if case_insensitive {
-            let this = self.as_bstr(ctx).to_lowercase();
-            let other = other.as_bstr(ctx).to_lowercase();
-            this.contains_str(other)
+            let this = self.as_bstr(ctx);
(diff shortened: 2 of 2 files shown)


plusvic:
Optimize case-insensitive string operations for ASCII inputs.

This avoids allocating lowercase copies for both operands in `icontains`, `istartswith`, `iendswith`, and `iequals` when both runtime strings are ASCII. For non-ASCII inputs the previous behavior is preserved.

PR #655, opened 2026-05-06
feat: add external modules support

1ndahous3:
Adds a first-class API for registering YARA-X modules from external crates, including support for callable functions accessible from YARA-X rules.

### Why
Currently, adding a YARA-X module requires forking the repository and modifying internal code. This change exposes a stable external API so third-party crates can ship their own modules without touching yara-x internals.

### Changes
- `YaraModule` public struct - name, proto descriptor, optional main function, optional `rust_module_name` for function lookup.
- `yara_x::inventory` re-export so external crates link against the exact same inventory version.
- `#[module_export(yara_x_crate = "yara_x")]` - new argument on the existing macro. When set, the macro generates fully-qualified type paths (`yara_x::Caller`, `yara_x::WasmExport`, etc.) and emits an unconditional `inventory::submit!`, making the macro usable outside yara-x itself.
- Public type aliases for macro-generated glue code (`ScanContext`, `Caller`, `WasmExport`, `WasmExportedFnN`) - all `#[doc(hidden)]` except `ScanContext` and `module_export` which appear in user-written code.
- `examples/custom-module-example` - a self-contained crate demonstrating registration, a `main_fn`, a callable `add(a, b)` function, `set_module_output` override, and a serialize/deserialize round-trip. Six integration tests cover all scenarios.
- Refactoring existing internal modules. Integrating them through a unified "inventory" mechanism.


plusvic:
I like this idea of allowing third-party creatse to inject their own modules without touching this repository. But now that we are working on it, I would try to make this method the one and only. Did you find any roadblock for using this method with the official modules as well?

This would mean dropping support for `linkme` and use `inventory` instead, but that's something that I had in mind anyways. This will also reduce complexity and as we have a unified way of defining modules, no matter if they are third-party or not.

1ndahous3:
@plusvic at this point, we (you in `main` and I in this request) have completely dropped "linkme" method. Do you want me to move the internal modules to a different crate, or is this sufficient?

plusvic:
> @plusvic at this point, we (you in `main` and I in this request) have completely dropped "linkme" method. Do you want me to move the internal modules to a different crate, or is this sufficient?

I still see the `custom-modules` feature and the use of `include!("add_modules.rs");` for inserting modules into `BUILTIN_MODULES`. My idea is removing the all mechanism altogether and use the new one as the official way to track the modules.

1ndahous3:
@plusvic I think it's done :)

docs/ModuleDeveloperGuide.md
```diff
@@ -25,6 +25,13 @@ words
 - [Adding dependencies](#adding-dependencies)
 - [Using enums](#using-enums)
     - [Inline enums](#inline-enums)
+- [Creating a module in an external crate](#creating-a-module-in-an-external-crate)
```
willdurand, line 28: Hey. I am curious: could we extend this to python bindings somehow?

plusvic: For modules that only export data using a protobuf, maybe. For modules that implement functions that would be a lot more complicated.

willdurand: If the module implements functions, etc. what is the issue exactly?  Is the way this external crate works not compatible with functions, etc. or is the problem caused by the python bindings? I guess what I am trying to understand is whether we could package our own module by using this external crate feature, even if that means we still need to build the Yara rust lib and the python bindings ourselves (that'd still be an improvement over maintaining an active fork).

plusvic: I think I misunderstood the original question. Do want to write a YARA module in Python? Or do you want to use your custom module written in Rust using the Python extension?

My prior answer was about the former: writing a YARA module in Python. However I guess that using a custom module written Rust is not straighforward neither, because the Python extension is currently linked against the yara_x crate, not your custom crate. Maybe you can add the crate implementing your custom module as a dependency of the `yara-x-py` crate and that's enough, but that must be tested.

willdurand: > I think I misunderstood the original question. Do want to write a YARA module in Python? Or do you want to use your custom module written in Rust using the Python extension?

Yeah, that's what I assumed, no worries, I could have been more verbose upfront. We're using yara in python, but our module is a rust module (currently in a fork, and implemented like the other built-in modules).  My question was about using a rust module with this new feature through python.

I'll wait until this lands, and then I'll try what you're suggesting.

PR #669, opened 2026-06-01
fix: rename binary release archives from `.gz` to `.tar.gz`

FranciscoPombal:
Release archives for non-Windows targets are gzip-compressed TAR files (`tar czf`) but were being named with a `.gz` extension, causing interoperability issues with tooling that infers format from the extension.

The artifact upload and release steps use glob patterns (`yara-x-*`) and are unaffected. Windows `.zip` builds are unaffected. Downstream consumers fetching assets by exact filename will need to update from `.gz` -> `.tar.gz`.


FranciscoPombal:
Any guidance concerning the Copilot-triggered CLA failure?

I can fully author the commit myself if that's what's needed.

Thanks.

PR #689, opened 2026-06-23
fix: don't derive header constraints from patterns with byte-transforming modifiers

nugxperience:
*Authored by Claude (Anthropic, Opus 4.8); submitted by @nugxperience.*

### Summary

A pattern using the `xor` (or `nocase`/`wide`/`base64`/`base64wide`) modifier, combined with a position anchor `$x at 0`, fails to match even when the pattern is present at offset 0.

Minimal repro — this rule does **not** match a file that begins with `"Hello"` XORed with key `0x01` (bytes `49 64 6d 6d 6e`), even though the match is exactly at offset 0:

```yara
rule t {
    strings:
        $x = "Hello" xor
    condition:
        $x at 0
}
```

The same failure occurs with `nocase` (when the data case differs from the literal), `wide`, `base64`, and `base64wide`. It only manifests at offset `0`; `$x at 4` and other offsets work correctly, and plain literals work correctly.

### Regression

This is a regression in **v1.18.0**. v1.16.0 and v1.17.0 are unaffected. It was introduced by the header-constraints optimization in #676 (`perf: optimize scans using header constraints`).

### Root cause

When a condition anchors a pattern at offset 0 (`$a at 0`), the compiler derives a "header constraint" — the bytes the file must begin with — and at scan time prunes the pattern early if the file header can't satisfy them (`lib/src/scanner/context.rs`).

The closure that supplies those bytes (`lib/src/compiler/mod.rs`) returned the pattern's literal text and **ignored the pattern's modifiers**:

```rust
Pattern::Text(lit) => Some(lit.text.as_bytes().to_vec()),
```

A header constraint is a *necessary condition* check: it's only valid when the bytes that must appear at offset 0 are exactly the literal text. That invariant holds for plain literals (and `ascii`/`fullword`, which don't change the bytes), but it is broken by modifiers that transform the on-disk bytes:

- `xor` — bytes are XORed with an unknown key
- `nocase` — any case variant is allowed
- `wide` — bytes are interleaved with zeroes
- `base64` / `base64wide` — bytes are base64-encoded

For these, the file starts with the *transformed* bytes, so `is_satisfied` returns false, the pattern is added to `disabled_patterns`, its atom hits at offset 0 are never verified, and `is_pat_match_at` returns false → the rule misses.

The regexp branch has the same latent issue for `wide`: it is lowered to a sub-pattern flag **after** the HIR is built, so it isn't reflected in `as_literal_bytes()`. (`nocase` is already baked into the regexp HIR at parse time, so its guard there is defensive; hex patterns accept no modifiers, so their guard is inert but kept for uniformity.)

### Fix

Return no prefix bytes (and therefore derive no header constraint) for any pattern carrying a byte-transforming modifier, in both the text and regexp/hex branches. Plain literals keep the optimization unchanged.

```rust
let non_literal_flags = PatternFlags::Xor
    | PatternFlags::Nocase
    | PatternFlags::Wide
    | PatternFlags::Base64
    | PatternFlags::Base64Wide;
let header_constraints = self.ir.header_constraints(|pat_idx| {
    let pat = &rule_patterns[pat_idx.as_usize()];
    match pat.pattern() {
        Pattern::Text(lit) => {
            if lit.flags.intersects(non_literal_flags) {
                None
            } else {
                Some(lit.text.as_bytes().to_vec())
            }
        }
        Pattern::Regexp(re) | Pattern::Hex(re) => {
            if re.flags.intersects(non_literal_flags) {
                None
            } else {
                re.hir.as_literal_bytes().map(|bytes| bytes.to_vec())
            }
        }
    }
});
```

### Testing

- Added regression cases to `header_constraints_optimization` covering `xor`/`nocase`/`wide`/`base64`/`base64wide` + `at 0`, a regexp `/Hello/ wide` + `at 0` case, a negative case (encoded form absent at offset 0 → correctly does not match), and a mixed-pattern case proving a modifier pattern doesn't contaminate a plain literal's constraint in the same rule.
- Full `yara-x` library test suite passes (303 tests, 0 failures), including the existing `header_constraints` and `header_constraints_optimization` tests.
- The CI checks pass locally: `cargo fmt --all --check` and `cargo clippy --tests --no-deps -- --deny clippy::all` are both clean.
- Manually verified against a from-source build: `xor $x at 0` now matches; `at 4`, plain `at 0`, and the negative case (literal absent at 0 → still correctly pruned) all behave as expected, confirming the optimization is preserved.

### Files changed

- `lib/src/compiler/mod.rs` — skip header-constraint derivation for byte-transforming modifiers.
- `lib/src/tests/mod.rs` — regression tests.

### DIFF
--- a/lib/src/compiler/ir/mod.rs
+++ b/lib/src/compiler/ir/mod.rs
@@ -40,7 +40,7 @@ use std::ops::{Add, Index};
 use std::rc::Rc;
 
 use bitflags::bitflags;
-use bstr::BString;
+use bstr::{BString, ByteSlice};
 use rustc_hash::FxHasher;
 use serde::{Deserialize, Serialize};
 
@@ -1007,9 +1007,22 @@ impl IR {
         result
     }
 
-    pub fn header_constraints(
+    /// Determines the constraints on the file header imposed by a rule condition.
+    ///
+    /// This function analyzes the rule's condition to determine whether it
+    /// restricts matching to files that start with a specific sequence of bytes.
+    ///
+    /// For example, the condition `uint32(0) == 0x464c457f and $a` requires that
+    /// the first 4 bytes of the file are `0x7f, 0x45, 0x4c, 0x46` (little-endian).
+    /// Similarly, `$a at 0` imposes a header constraint if `$a` has a known
+    /// constant prefix.
+    ///
+    /// In contrast, the condition `uint32(0) == 0x464c457f or $a` does not impose
+    /// a header constraint, since the use of `or` allows files with a different
+    /// header to also match.
+    pub fn header_constraints<'a>(
         &self,
-        pattern_prefix_lookup: impl Fn(PatternIdx) -> Option<Vec<u8>>,
+        pattern_lookup: impl Fn(PatternIdx) -> &'a Pattern,
     ) -> HeaderConstraint {
         let mut constrained_bytes = BTreeMap::new();
         let mut unsatisfiable = false;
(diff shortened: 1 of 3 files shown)


plusvic:
Good catch @nugxperience, thanks for this PR.

Issue #319, opened 2025-03-06
Feature Request: Add data and plaintext to Match results in Python and Go libraries


zdiff:
I think it would be useful to print matching patterns for matches when using the Go and Python YARA-X libraries.

plusvic:
Handling matching strings can be tricky, especially when dealing with extremely large strings in regular expressions or hex patterns.

In the Rust API, matching strings are directly backed by the scanned data itself—no copies are made. As a result, exposing matching strings is a zero-cost operation.

However, this approach is not feasible in the Go and Python APIs. To expose matching strings in these languages, copies would have to be made, which introduces limitations (e.g., copying multi-gigabyte matching strings is not feasible). To avoid this, the Go and Python APIs provide only the starting offset and length of each match. It is up to the user to retrieve the matching string from the original scanned data using this information.

PR #692, opened 2026-07-01
feat: typing on match and return data

tlansec:
I realized today when writing some code with yara_x's Python API that typing information wasn't exposed for matching patterns:

image

This PR also tries to fix that.

I also realized that the matched data isn't currently included, and when using the `.scan_file()` method you have to re-read the file back to get matched bytes back.

This PR also tries to fix that.

### DIFF
--- a/py/src/lib.rs
+++ b/py/src/lib.rs
@@ -1190,8 +1190,6 @@ struct Match {
     /// For patterns that have the `xor` modifier, contains the XOR key that
     /// applied to matching data. For any other pattern will be `None`.
     xor_key: Option<u8>,
-    /// The matched data.
-    data: Py<PyBytes>,
 }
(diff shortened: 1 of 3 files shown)


plusvic:
The `data` field wasn't included in the `Match` structure on purpose. The reason can be found in #319.

I can merge the rest of the changes.

tlansec:
Ah - I didn't know that.

Would you like me to remove that or would you prefer to do it?

plusvic:
> Ah - I didn't know that.
> 
> Would you like me to remove that or would you prefer to do it?

Yes please, do it and I'll merge it.

tlansec:
I made the change and removed the match data related changes, keeping this PR to just adding the type hints.

PR #704, opened 2026-07-12
feat: add --ignore-invalid-rules to skip rules that fail to compile

### LINKED ISSUE #244: Ignore errors on rule compile

paul-atx:
Adds a `--skip-invalid-rules` flag to the `compile` and `scan` commands. When set, rules that fail to compile are skipped and compilation continues with the valid ones, instead of the whole run aborting on the first error. This helps when running large third-party rulesets that aren't fully compatible with YARA-X.

Unlike #307, this keeps per-rule granularity: the compiler already discards only the individual rules that fail (via its snapshot/rollback), so the valid rules contained in a file that also has a broken rule are kept, rather than the whole file being dropped.

Every compilation error is still printed, and `yr compile --skip-invalid-rules` still exits with a non-zero code when errors were found — the valid rules are written to the output first. This preserves the existing "errors were found" exit semantics rather than changing them.

The flag is off by default, so current behavior is unchanged.

Closes #244.


plusvic:
I've changed `--skip-invalid-rules` to `--ignore-invalid-rules` for consistency with `--ignore-module`. Also, when `--ignore-invalid-rules`  the returned error code is always zero, unless some other error occurs that is not related to compilation. I think that user's won't expect that the CLI returns a non-zero error code when told explicitly that compilation errors should be ignored.

paul-atx:
Thanks! Agreed on the exit code. With an explicit `--ignore-invalid-rules`, exit zero is the intuitive behavior. My non-zero version was leaning too hard on the earlier concern about the default behavior changing. The rename and the "list ignored rules with reasons" addition are nice improvements.

PR #711, opened 2026-07-15
feat: Expose ignore_invalid_rules in python API.

wxsBSD:
This commit exposes the new feature of ignoring invalid rules during
compilation. It is set to False by default but if turned on than errors are not
raised on compilation failure and any ignored rules will be returned when
calling the `Compiler::ignored_rules()` method. It returns a list of IgnoredRule
objects which have two properties: name and reason.

### DIFF
--- a/py/src/lib.rs
+++ b/py/src/lib.rs
@@ -437,6 +437,16 @@ impl Module {
     }
 }
 
+/// Structure that represents invalid rules by the compiler. See
+/// [`Compiler::ignore_invalid_rules`].
+#[pyclass]
+struct IgnoredRule {
+    #[pyo3(get)]
+    name: String,
+    #[pyo3(get)]
+    reason: String,
+}
+
 /// Returns the names of the supported modules.
 ///
 /// These are the modules that can be used in `import` statements in your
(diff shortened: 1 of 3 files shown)


wxsBSD:
<s>I'm thinking a better way to do this is to expose the `IgnoredRuleReason` enum information on each of the `IgnoredRule` objects we return. This way users can decide to treat a `IgnoredRuleReason::CompileError` differently from a `IgnoredRuleReason::IgnoredModule` if they desire. Right now the only way to tell is to look at the reason property on the object and see if it starts with `error: ` or not.

I didn't implement it this way at first because I literally based it off the way you handle things in the cli. If you think exposing the `IgnoredRuleReason` in a programatic fashion is the right thing to do let me know and I'll implement it!</s>

I added support for checking the reason via an enum that we expose in the python API. I tried to re-export the one from yara-x directly but doing it with proper pyo3 support was difficult (impossible maybe?). As such, I just define my own enum and we should always be kept in sync thanks to rust analyzer complaining if a new one is added in yara-x but not in the python API.

plusvic (approved):

PR #722, opened 2026-07-27
feat(lnk): parse ShellItems and malformed sections

MichaelTheSynthCat:
Improvements in LNK module:

- parse Shell Items
- parse sections in lnk files with invalid section sizes (abused by CVE-2010-2568 exploits)

---

The Windows Shell Item format isn't officially standardized by Microsoft and therefore community-made documents are the best available source of truth:

- [libfwsi](https://github.com/libyal/libfwsi/blob/main/documentation/Windows%20Shell%20Item%20format.asciidoc)
- [LnkParse3](https://github.com/Matmaus/LnkParse3/tree/master) - referenced by @plusvic in the TODO comment.

Shell item types differ in their internal structure while some parts of them are not fully understood but some of them contain encoded valuable data (filepaths, URLs, which are usually UTF-16 encoded) that could have their own field in the lnk module proto output to improve user experience. I would like to open a discussion especially about the output of the LNK module regarding the shell items:

 - which shell item types should we recognize
 - which fields of a certain shell item type could we expose in LNK proto and how to include them there

I have implemented  yet parsing of some basic shell item types and exposed the most valuable fields in the proto. Maybe we could just copy most of the parsing of extracted values from shell items as it's done in [LnkParse3](https://github.com/Matmaus/LnkParse3/tree/master)?


lib/src/modules/protos/lnk.proto
```diff
@@ -115,6 +115,60 @@ message Lnk {
...
+// Microsoft. The type-specific fields below therefore follow the
+// reverse-engineered "Windows Shell Item format" documentation by Joachim Metz
+// (libfwsi). Which fields are populated depends on the shell item type (see
+// `item_type`); the raw `data` field is always populated.
+//
+// The fields are ordered by the `item_type` (range) they are extracted from.
+message ShellItem {
+  // Shell item class type indicator: the byte that identifies the kind of
+  // shell item (e.g. 0x00 control panel CPL file, 0x1F root folder).
+  optional uint32 item_type = 1;
```
plusvic: `item_type` should be an enum.

plusvic (changes requested):

lib/src/modules/protos/lnk.proto
MichaelTheSynthCat: Good idea, now it looks more human-readable. I have still left `item_type_code = 2 [(yara.field_options).fmt = "x"]` in case an unknown item type is encountered.

PR #725, opened 2026-07-30
feat: add HydraDragon/androguard modules

Siradankullanici:
https://github.com/HydraDragonAntivirus/yara-x
This code inspired from androguard come from Koodus and compatiable with them. Only one extra module exists called hydradragon. It shouldn't compiled as default module but as optional module since it requires sending json data to YARA-X with this project: https://github.com/HydraDragonAntivirus/HydraDragonAV-Mobile so we need better implamanation without requring this mobile antivirus sends to data or keep only hydradragon as send json data.
This also contains daachorse improvements without using any GPLv2 code from ClamAV.

Only known problem is daachorse didn't integrated prefiltering yet.


plusvic:
There are a few issues with this PR:

* The PR is mixing multiple features together. In cases like this, when you want to merge two different modules and the prefilter feature, it's better to send three independent PRs: one for the prefilter feature alone, one for the androguard module, and another one for the hydradragon module. Each PR should cover one specific feature, using the same PR for multiple independent features makes it harder to review.

* I've done some benchmarking with the prefilter feature and didn't find any performance boost, quite the contrary, the prefilter feature is slowing down YARA-X. According to the documentation, the prefilter is used "only when all atoms fit in 64 total bytes", so I used this simple rule in my tests:

```
rule singlepattern {
	strings:
		$str = "foobarbaz"
	condition:
		$str
}
```

The results were:

```
|-----------------------------------------------------------------------|
| Legacy YARA          | YARA-X (prefilter)   | YARA-X (no prefilter)   | 
|-----------------------------------------------------------------------|
| 2710.89 ms ± 3.60    | 3727.17 ms ± 9.98    | 639.89 ms ± 10.11.      | 
|-----------------------------------------------------------------------|
```

With prefilter enabled YARA-X is around 5x slower, even slower than legacy YARA. Given that the use of the prefilter is limited to rules with only a few short patterns, the prefilter will be used in conjunction with the Teddy algorithm, not with Aho-Corasick. The Teddy algorithm is already very fast, and using the prefilter with it only makes scanning slower, not faster.

If you have performed other tests that reach a different conclusion please let me know the exact steps that I must follow to reproduce your results.

 * I prefer not merging the Androguard module at this point. This module is owned by the Koodous project and I don't know which are their plans regarding this module or its integration with YARA-X. 

* Hydragon is not a consolidated standard in the security industry yet. Most YARA-X users won't be using a module that depends on an output produced by Hydragon. As YARA-X already supports the integration with modules that are developed in external repositories, you can keep working on this module in your own repository and distribute it together with YARA-X in your product, but I see little benefit from merging this code in the YARA-X's repository at this point.

Siradankullanici:
I will create new pull request only for prefilter usage and actually androguard module is has  nothing with real androguard because it uses hydradragon backend to get json data. Only similarity is function names.
Also slowness come from wrong usage and wrong algorithm since ClamAV's algorithm more faster I will keep from there at daachorse but I can't create pull request so I need new from scratch algorithm.

Edit: I probably can't create pr due to policies of daachorse.

Siradankullanici:
I reverted code to which contains GPLv2 and my own code with extra optimization. It become from 9%-25% at very small file now 30% with passing all tests but at bigger more than 45%

Issue #737, opened 2026-08-20
Scan performance regression (~4-6x) since v1.18.0


1ndahous3:
While scanning a large ruleset (thousands of rules) against a directory full of small files, I noticed scans got a lot slower going from v1.17.0 to v1.18.0. Single-threaded (`-p 1`) it's 4-6x slower depending on ruleset shape. Bisecting `v1.17.0..v1.18.0` points straight at b0d3360b, #676.

**Root cause:** `search_for_patterns()` used to check filesize bounds lazily, only for a pattern whose atom actually got hit during the AC search. #676 changed this to two loops over **every pattern in the entire ruleset**, run on **every single scanned file**, before the AC search even starts. For a big ruleset scanning lots of small files, that eager pass dominates runtime even when nothing ever matches.

I noticed 8f7ac9f4 / f28330e1 already improve this (iterating the constraint maps directly instead of doing a lookup per pattern id), but it doesn't fully close the gap - see numbers below.

**Repro**: generated test rules + target scan files attached. Roughly: ~5200 rules / ~42000 patterns, one rule that forces a real string search, the rest each guarded by a `uint32(0) == X` header check (so most patterns are "constrained" but still short-circuit on their own) - scanned against a few thousand tiny files that never match anything.

### Stats

```sh
time ./yr scan -w -r -p 1 rules.yar scan-files
```

| build | time | vs v1.17.0 |
|---|---|---|
| v1.17.0 | ~1.71s | baseline |
| b0d3360b (#676) | ~10.16s | 5.9x |
| 8f7ac9f4 / f28330e1 (unreleased, on `main`) | ~7.18s | 4.2x |
| 8f7ac9f4 / f28330e1 + #738 patch | ~5.33s | 3.1x |

The unreleased fix cuts it by ~30%, but it's still over 4x slower than v1.17.0 in this scenario, because a large share of the patterns here do carry a header constraint, so "only iterate constrained patterns" doesn't shrink the work much - and the check is still eager (runs for every constrained pattern on every file regardless of whether it'll save anything), unlike the old lazy, atom-hit-triggered check.

[scan-files.zip](https://github.com/user-attachments/files/31271362/scan-files.zip) 

[rules.zip](https://github.com/user-attachments/files/31271379/rules.zip)

1ndahous3:
I won some perf with #738 fix (added to table) but I don't think it's enough...

plusvic:
[...]

PR #738, opened 2026-08-20
perf: use a `Vec<bool>` for `MatchTracker::disabled_patterns`

1ndahous3:
`MatchTracker::disabled_patterns` gets cleared and rebuilt on every single scanned file, and for rulesets where a lot of patterns carry a filesize/header constraint that's a lot of entries touched per file. `PatternId` is already a dense `0..num_patterns` index, so a plain `Vec<bool>` sized once when the `ScanContext` is created is cheaper here than any set type - `.clear()` becomes `.fill(false)` (a memset), and insert/contains become plain indexing.

Benchmark (repro from #737):

| build | time | vs upstream |
|---|---|---|
| upstream | ~7.18s | baseline |
| \+ this patch | ~5.33s | 26% faster |

### DIFF
--- a/lib/src/scanner/context.rs
+++ b/lib/src/scanner/context.rs
@@ -21,7 +21,7 @@ use bstr::{BString, ByteSlice};
 use indexmap::IndexMap;
 use protobuf::{MessageDyn, MessageFull};
 use regex_automata::meta::Regex;
-use rustc_hash::{FxHashMap, FxHashSet};
+use rustc_hash::FxHashMap;
 use smallvec::SmallVec;
 
 use crate::compiler::{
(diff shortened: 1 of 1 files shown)


ikrivosheev:
I think is not the best fix option. Better is to use vector of bits to not consuming a lot memory...

1ndahous3:
@ikrivosheev `bitvec` is slower than `Vec<bool>`, confirmed for this case.

ikrivosheev:
@1ndahous3 but is better with memory. For large rules set it can be sensitive

plusvic:
This has been addressed in a different way in #741. 

For the time being we are not using a bit vector because it introduces a O(N) vector clearing operation with every scan, which could start to being noticeable when the number of patterns is large enough. Now that the percentage of disabled patterns should be reduced, using a vector instead of a hashmap makes less sense than before.

PR #743, opened 2026-08-24
perf: apply file size and file header constraints only to regular expressions

1ndahous3:
Additional fix to patch 529a207d51e677b18fa95696783531a017172b2f, made as part of #737


plusvic:
In def4461 I've refactored the code and to have a single source of truth for constraints and this change is no longer necessary.

PR #748, opened 2026-08-27
feat: add debug command for selected atoms

matt-acsc:
Adds a yr debug atoms command for displaying the final atoms selected by YARA-X for each rule string.

The command is available behind the existing debug-cmd feature and outputs selected atoms grouped by rule and pattern.

## Example

```text
rule test
  $a
    41424344
  $b
    01020304
```
This is intended for tooling and debugging use cases that need visibility into YARA-X's final selected atoms.

One motivating use case is tooling that needs to consume YARA-X's selected atoms externally, for example to feed them into an indexed prefilter/search system such as BigGrep before performing full rule evaluation.

Tests have been added for the new command and the workspace test suite passes.

### DIFF
--- a/cli/src/commands/debug.rs
+++ b/cli/src/commands/debug.rs
@@ -25,6 +25,7 @@ pub fn atoms() -> Command {
                 .help("Path to YARA source file")
                 .value_parser(value_parser!(PathBuf)),
         )
+        .arg(arg!(--json).help("Print output in JSON format"))
 }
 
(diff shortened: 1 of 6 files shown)


plusvic:
The idea is great, in fact I've thinking about adding this for some other reasons, but the implementation could be done in a more efficent way that doesn't require allocating vectors. Also an API for iterarting over the rules and their patterns alrady exists, and could be used.

I will make my changes to this PR.

PR #752, opened 2026-08-28
perf(wasm): cache constant input header reads

king-tero:
## Summary

Cache the first 128 bytes of contiguous input in WASM memory and emit direct
guest loads for constant unsigned 16- and 32-bit integer reads within that
prefix. This avoids repeated host calls for common header checks such as
uint16(0) and uint32(0x3c).

Dynamic offsets, signed reads, reads outside the cached prefix, short inputs,
and block scans retain the existing host fallback and undefined-value
semantics. Scanner reset clears the cached length so a short scan cannot see
bytes left by a previous longer scan.

The reserved memory moves the matching-rules bitmap, so the compiled-rules
serialization version is intentionally bumped from 6 to 7. Old and new
readers reject the opposite package version explicitly instead of interpreting
different WASM memory offsets.

## Results

Measured from current main at 24cd993c on a Linux c4-standard-32 using
release-lto, target-cpu=native, glibc transparent huge pages, and 11
interleaved pairs.

| Workload | Before | After | Paired mean | Wins | 95% CI |
|---|---:|---:|---:|---:|---:|
| Forge Core, 10,000 x 4 KiB, t1 | 1,088.8 files/s | 1,156.8 files/s | **+6.039%** | 11/11 | +5.568% to +6.449% |
| Forge Core, 100,000 x 4 KiB, t32 | 25,755 files/s | 27,293 files/s | **+6.089%** | 11/11 | +5.438% to +6.777% |

A same-runtime package-only comparison isolates the generated-code change at
**+5.411%** on the 100,000-file t32 workload, with 11/11 wins and a 95%
bootstrap confidence interval of +4.925% to +5.958%.

Controls:

- Forge Core, 256 MiB, t1: +0.192%; CI -1.234% to +1.595%
- Forge Full modules, t4: -0.380%; CI -1.507% to +0.779%
- regex-heavy, 1 GiB, t1: +0.387%; CI -0.029% to +0.800%

Sorted Forge Full output is identical. A same-runtime Core comparison between
packages with and without direct cached loads is also identical.

## Validation

- cargo test --no-default-features --features=default-modules,magic-module
- cargo test --features=magic-module,rules-profiling
- 52 explicit CLI tests
- focused little-/big-endian, cache-boundary, scanner-reuse, block-scan, and serialization tests
- cargo clippy --tests --no-deps -- --deny clippy::all
- cargo fmt --all -- --check
- git diff --check

### DIFF
--- a/lib/src/compiler/emit.rs
+++ b/lib/src/compiler/emit.rs
@@ -21,6 +21,7 @@ use walrus::ir::{
 };
 use walrus::{FunctionId, InstrSeqBuilder, ValType};
 
+use crate::compiler::ir::FuncCall;
 use crate::compiler::ir::{
     Expr, ExprId, ForIn, ForOf, IR, Iterable, MatchAnchor, PatternIdx,
     Quantifier,
@@ -39,6 +40,7 @@ use crate::wasm;
 use crate::wasm::builder::WasmModuleBuilder;
 use crate::wasm::string::RuntimeString;
 use crate::wasm::{
+    CACHED_HEADER_BASE, CACHED_HEADER_CAPACITY, CACHED_HEADER_LEN_BASE,
     LOOKUP_INDEXES_END, LOOKUP_INDEXES_START, MATCHING_RULES_BITMAP_BASE,
     VARS_STACK_START, WasmSymbols,
 };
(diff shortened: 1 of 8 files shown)


plusvic:
This is too much added complexity for a low performance gain.

PR #755, opened 2026-08-30
perf(regex): swap PikeVM thread buffers by reference

king-tero:
## Summary

Borrow the PikeVM thread buffers once and swap their local references after
each input byte, instead of swapping both `Vec` headers.

The active and next thread sets, instruction decoder, matching logic, bytecode
format, and public API are unchanged. This only removes repeated pointer,
length, and capacity movement in the PikeVM hot loop.

## Results

Measured from `main` at `02a0ddda` on a dedicated 32-vCPU Linux host with
Rust 1.98, release LTO, `target-cpu=native`, clean sequential builds in the
same absolute path, and byte-identical same-inode null arms.

| Workload | Result | 95% CI |
|---|---:|---:|
| Regex match-positive, 8 x 16 MiB, 1 thread | **+56.502% (12/12)** | **+54.931% to +57.710%** |
| Forge Core, 100k x 4 KiB, 32 threads | **+1.447% (12/12)** | **+1.250% to +1.666%** |
| Forge Core, 256 MiB, 1 thread | +0.075% (9/12) | -0.475% to +0.534% |
| Forge Full/modules, 150 files, 4 threads | +0.706% (7/12) | -0.486% to +1.991% |
| Regex no-match, 1 GiB, 1 thread | -0.107% across 5 sessions | -0.582% to +0.368% |

Regex throughput increases from 53.3 to 83.4 MB/s in the fresh gate. A prior
three-session clean-build confirmation on the same `main` measured +58.802%
(54/54, between-session CI +58.496% to +59.108%).

Sorted NDJSON output is byte-identical for all 8 regex records and all 100,000
Core records.

## Validation

- `cargo +1.98.0 test --locked --workspace`
- `cargo +1.98.0 test --locked --workspace --no-default-features`
- 53 explicit CLI tests
- `cargo +1.98.0 clippy --locked --workspace --tests --no-deps -- --deny clippy::all`
- `cargo +1.98.0 fmt --all -- --check`
- `git diff --check`

### DIFF
--- a/lib/src/re/thompson/pikevm.rs
+++ b/lib/src/re/thompson/pikevm.rs
@@ -171,17 +171,23 @@ impl<'r> PikeVM<'r> {
         // called.
         debug_assert!(self.threads.is_empty());
 
+        let code = self.code;
+        let scan_limit = self.scan_limit as usize;
+        let cache = &mut self.cache;
+        let (mut threads, mut next_threads) =
+            (&mut self.threads, &mut self.next_threads);
+
         epsilon_closure(
-            self.code,
+            code,
             start,
             0,
             curr_byte,
             bck_input.next(),
-            &mut self.cache,
-            &mut self.threads,
+            cache,
+            threads,
         );
 
-        while !self.threads.is_empty() {
+        while !threads.is_empty() {
             let mut next_byte = fwd_input.next();
             // When there is only a single active thread in the VM (that is,
             // `self.threads.len() == 1`), we can optimize execution by
@@ -194,12 +200,12 @@ impl<'r> PikeVM<'r> {
             // desynchronize and bypass other threads matching at different
             // positions or branches. It's safe to set decode_literal_runs
             // always to false, it will simply disable the optimization.
-            let decode_literal_runs = self.threads.len() == 1;
+            let decode_literal_runs = threads.len() == 1;
 
-            for (ip, rep_count) in self.threads.iter() {
+            for (ip, rep_count) in threads.iter() {
                 let ip = *ip as usize;
                 let (instr, mut instr_size) = InstrParser::decode_instr(
-                    unsafe { self.code.get_unchecked(ip..) },
+                    unsafe { code.get_unchecked(ip..) },
                     decode_literal_runs,
                 );
 
@@ -285,13 +291,13 @@ impl<'r> PikeVM<'r> {
 
                 if is_match {
                     epsilon_closure(
-                        self.code,
+                        code,
                         C::from(ip + instr_size),
                         *rep_count,
                         next_byte,
                         curr_byte,
-                        &mut self.cache,
-                        &mut self.next_threads,
+                        cache,
+                        next_threads,
                     );
                 }
             }
@@ -299,11 +305,11 @@ impl<'r> PikeVM<'r> {
             curr_byte = next_byte;
             current_pos += 1;
 
-            mem::swap(&mut self.threads, &mut self.next_threads);
-            self.next_threads.clear();
+            mem::swap(&mut threads, &mut next_threads);
+            next_threads.clear();
 
-            if current_pos >= self.scan_limit as usize {
-                self.threads.clear();
+            if current_pos >= scan_limit {
+                threads.clear();
                 break;
             }
         }


plusvic:
I would like to have more details about how the performance was measured and how to reproduce the 50% speed up claimed in the PR. In my own tests I can observe a 3-4% improvement with a 100% PikeVM-bound workload.

My own conclusions are:

* Code change is sound and worth merging: The change in `PikeVM::try_match_impl` is small (7 lines), clean, preserves all invariants (`self.threads` and `self.next_threads` are always empty on exit), and produces measurably better machine code (replaces a 48-byte Vec-header swap + store-to-load forwarding hazard with register moves, and keeps self.code and self.scan_limit in registers). 

* Real-world speedup is modest (~0–1% overall, ~3–4% on pure PikeVM loops): On general rulesets, YARA-X spends most time in atom matching (Aho-Corasick / Teddy) and FastVM, so end-to-end impact is ~0–1%. 

* Even on a synthetic 100% PikeVM-bound benchmark compiled with `--profile release-lto and RUSTFLAGS="-C target-cpu=native"`, the actual speedup is ~3.5–3.8% (9/12 wins). The +56.5% figure only happens on a contrived microbenchmark where a specific x86_64 store-to-load forwarding stall dominates a minimal 1-thread loop.

plusvic (approved):

king-tero:
Thanks for the review and merge. I agree that the 56.5% figure must not be presented as a general YARA-X or even a typical PikeVM-bound speedup. It is specific to the synthetic input, executable pair and machine used in that experiment.

The exact rules and data are the same as the [generator and hashes I posted on #757](https://github.com/VirusTotal/yara-x/pull/757#issuecomment-5758913395): one rule with 200 patterns of the form `/YRX000[A-Za-z0-9_]{4,12}PERF/`, through `YRX199`, and `any of them`; eight separate 16 MiB files containing identical repetitions of `YRX000ABCDEFGHPERF\n`. They are not eight diverse real-world samples. Normal scanning was used, without `--fast-scan`.

For **this** PR, the source pair was `02a0ddda2f1d018948492c0ac9a822da84752cbc` versus `972b87d6a5fa4207811f00bf1af2fba13ebade7d`, without #757 or other pending patches. Both builds used Rust 1.98.0 on an Intel Xeon Platinum 8581C Linux VM, with:

```sh
RUSTFLAGS='-C target-cpu=native' cargo +1.98.0 build --locked \
  --profile release-lto -p yara-x-cli --features native-code-serialization
```

They were clean sequential builds at the same checkout/target path, saving each executable before rebuilding. After running the linked generator, compile the rules once with the baseline, then use the same package for both executables:

```sh
./yr-baseline compile --output regex-heavy.yarc regex-heavy.yar
taskset -c 12 ./yr-baseline scan --compiled-rules --disable-console-logs \
  --threads 1 --scan-list regex-heavy.yarc files.list >/dev/null
taskset -c 12 ./yr-candidate scan --compiled-rules --disable-console-logs \
  --threads 1 --scan-list regex-heavy.yarc files.list >/dev/null
```

The quoted +56.502% came from 12 three-arm rounds after two warmups per arm, rotating all six orders. The two baseline arms invoked the **same executable path/inode** twice. The per-round calculation was `(mean(baseline times) / candidate time - 1) * 100`; the reported interval for that row is a within-session bootstrap, not cross-machine evidence. Median elapsed times were 2.517/2.516 s for the baseline arms and 1.609 s for the candidate (128 MiB total, about 53.3 → 83.4 MB/s). The archived package SHA-256 is `392773b8f14a1da65b11716df07a6faab855cf0d4d7dc3b48a1d5858300c7dad`.

I checked that the archived binaries and raw record still match their recorded hashes; this is the original measurement, not a new post-merge revalidation. Our representative controls were much smaller: +1.447% Core-small, +0.075% Core-large, and +0.706% Full with its interval crossing zero. None supports advertising a 50% general improvement. Your 3–4% PikeVM and ~0–1% overall measurements are the more appropriate expectations to communicate, rather than extrapolating from this narrow microbenchmark.

The important justification is the small, invariant-preserving change and the better generated code you identified. In future submissions I will label synthetic workloads explicitly, include their reproducer up front, and distinguish their gains from representative end-to-end results.

PR #757, opened 2026-08-30
perf(regex): fast-path common PikeVM instructions

king-tero:
## Summary

Fast-path `ClassBitmap` and `RepeatGreedy` before PikeVM's generic
extended-opcode dispatch.

These two instructions dominate prefixed dispatches in the measured
match-positive regex workload. All other instructions continue through the
existing decoder. The matching algorithm, bytecode format, serialized rules,
and public API are unchanged.

## Results

Measured from current `main` at `02a0ddda` with Rust 1.98, release LTO,
`target-cpu=native`, clean sequential builds in the same absolute path, and
byte-identical same-inode null arms on a dedicated 32-vCPU Linux host. Both
binaries scan the same compiled packages.

| Workload | Result | 95% CI |
|---|---:|---:|
| Regex match-positive, 8 x 16 MiB, 1 thread | **+6.283% across 3 sessions (36/36)** | **+5.295% to +7.271%** |
| Forge Core, 100k x 4 KiB, 32 threads | +0.294% (10/12) | +0.128% to +0.464% |
| Forge Core, 256 MiB, 1 thread | -0.311% | -0.890% to +0.194% |
| Forge Full/modules, 150 files, 4 threads | -0.108% across 5 sessions | -0.774% to +0.559% |
| Regex no-match, 1 GiB, 1 thread | -0.125% across 5 sessions | -0.686% to +0.436% |
| Literal no-match, 1 GiB, 1 thread | -0.080% | -0.872% to +0.669% |

Regex throughput increases from about 52.8 to 55.9 MB/s in the fresh gate. A
prior 31-round confirmation on the same `main` measured +6.978% (31/31, 95%
CI +6.745% to +7.213%). Sorted NDJSON output is byte-identical for all 8 regex
records and all 100,000 Core records. The optimized binary file adds 864
bytes.

## Validation

- `cargo +1.98.0 test --locked --workspace`
- `cargo +1.98.0 test --locked --workspace --no-default-features`
- `cargo +1.98.0 test --locked -p yara-x-cli`
- `cargo +1.98.0 clippy --locked --workspace --tests --no-deps -- --deny clippy::all`
- `cargo +1.98.0 fmt --all -- --check`
- `git diff --check`

### DIFF
--- a/lib/src/re/thompson/instr.rs
+++ b/lib/src/re/thompson/instr.rs
@@ -349,6 +349,35 @@ impl<'a> InstrParser<'a> {
         code: &[u8],
         decode_literal_runs: bool,
     ) -> (Instr<'_>, usize) {
+        if code.len() >= 2 && code[0] == OPCODE_PREFIX {
+            match code[1] {
+                Instr::CLASS_BITMAP if code.len() >= 34 => {
+                    let bitmap = unsafe { code.get_unchecked(2..34) };
+                    return (Instr::ClassBitmap(ClassBitmap(bitmap)), 34);
+                }
+                Instr::REPEAT_GREEDY
+                    if code.len()
+                        >= 2 + size_of::<Offset>()
+                            + size_of::<u32>()
+                            + size_of::<u32>() =>
+                {
+                    let offset = Self::decode_offset(&code[2..]);
+                    let min =
+                        Self::decode_u32(&code[2 + size_of::<Offset>()..]);
+                    let max = Self::decode_u32(
+                        &code[2 + size_of::<Offset>() + size_of::<u32>()..],
+                    );
+                    return (
+                        Instr::RepeatGreedy { offset, min, max },
+                        2 + size_of::<Offset>()
+                            + size_of::<u32>()
+                            + size_of::<u32>(),
+                    );
+                }
+                _ => {}
+            }
+        }
+
         match code[..] {
             [OPCODE_PREFIX, Instr::ANY_BYTE, ..] => (Instr::AnyByte, 2),
             [OPCODE_PREFIX, Instr::MASKED_BYTE, byte, mask, ..] => {


plusvic:
I've done my own measurements and didn't notice a 5-7% performance gain as claimed. In the best case I've found the gain is less than 2%. Can you provide some concrete examples (rules and files to be scanned) where a 5-7% performance gain is obtained?

king-tero:
Thanks for checking this. The 5–7% result is from a **synthetic, match-dense regex benchmark**, not a representative Forge corpus. I should have made that explicit and included the reproducer. The Forge controls in the PR were approximately neutral, so your broader results below 2% are not contradicted by my measurements.

The exact rule contains 200 patterns (`$r000` through `$r199`), each of this form:

```yara
$r000 = /YRX000[A-Za-z0-9_]{4,12}PERF/
```

The condition is `any of them`. The input is `YRX000ABCDEFGHPERF\n` repeated to 16 MiB, stored in eight separate files with **identical contents**, not eight different real-world samples. This exercises the four-range byte class and bounded greedy repetition heavily. Scanning uses normal mode, **without `--fast-scan`**.

This Python script regenerates the exact rule and input bytes (run in an empty benchmark directory):

```python
from pathlib import Path
import hashlib

rules = "rule regex_heavy {\n  strings:\n" + "".join(
    f"    $r{i:03d} = /YRX{i:03d}[A-Za-z0-9_]{{4,12}}PERF/\n"
    for i in range(200)
) + "  condition:\n    any of them\n}\n"
Path("regex-heavy.yar").write_bytes(rules.encode("ascii"))
seed = b"YRX000ABCDEFGHPERF\n"
size = 16 * 1024 * 1024
data = (seed * ((size + len(seed) - 1) // len(seed)))[:size]
paths = []
for i in range(8):
    path = Path(f"match-dense-regex-{i:02d}.bin").resolve()
    path.write_bytes(data)
    paths.append(path)
Path("files.list").write_text("".join(f"{p}\n" for p in paths))
print("rules:", hashlib.sha256(rules.encode("ascii")).hexdigest())
print("each input:", hashlib.sha256(data).hexdigest())
```

Expected SHA-256: rules `55a1f19d7e6139c8f224330f1c4891836aafb2e0ccd044cbe6f761ed670b69ac`; each input `7a8195ae2687afd69c220ca58f7503db7c1be6b97a13698a447b935d675715ad`.

The measured source pair is parent `02a0ddda2f1d018948492c0ac9a822da84752cbc` versus this PR's `516a8ed58f4575ede46aedce1274c17d52c5f622`, without other pending patches. Hardware is an Intel Xeon Platinum 8581C Linux VM. The build command for each revision was:

```sh
RUSTFLAGS='-C target-cpu=native' cargo +1.98.0 build --locked \
  --profile release-lto -p yara-x-cli --features native-code-serialization
```

The original builds were clean and sequential at the same checkout/target path, preserving each executable before cleaning and building the other revision. With those executables named `yr-baseline` and `yr-candidate`, compile once and scan the **same** package:

```sh
./yr-baseline compile --output regex-heavy.yarc regex-heavy.yar
taskset -c 0 ./yr-baseline scan --compiled-rules --disable-console-logs \
  --threads 1 --scan-list regex-heavy.yarc files.list >/dev/null
taskset -c 0 ./yr-candidate scan --compiled-rules --disable-console-logs \
  --threads 1 --scan-list regex-heavy.yarc files.list >/dev/null
```

Warm both executables first, time whole scan commands (not compilation), and interleave/rotate their order. The original three session means were +5.866%, +6.658%, and +6.325% throughput, using 12 three-arm rounds per session and the mean of two baseline runs as the reference; +6.283% overall, 95% between-session CI +5.295% to +7.271%. Those original null runs used the same baseline executable path/inode twice, not independently built baselines.

I have verified that the generator reproduces the original input hashes and that recompiling with the archived baseline recreates the original package byte-for-byte. Baseline and candidate also return identical NDJSON and diagnostics for all eight files.

I also reran the archived executable pair today with those regenerated inputs: **+6.028% throughput**, 95% between-session CI **+5.667% to +6.389%**, across three separate harness invocations, six fully order-balanced three-arm rounds each (18/18 positive). This time the null was a separate physical copy of the baseline: -0.001%, CI -0.895% to +0.894%. No allocator preload, PGO or BOLT was used. This is a reproduction of the original executable pair, **not a fresh build or revalidation after today's merges**.

I do not currently have a representative real-world corpus demonstrating a 5–7% gain for this patch. This example establishes the targeted case, not a general scanning-speed improvement; whether that narrower benefit warrants the additional dispatch cases is a separate maintenance tradeoff.

plusvic:
I've tested with the rules and files generated by the Python script and don't observe a relevant improvement on performance. The gain is too small to be noticeable in real-life cases across all platforms.

PR #765, opened 2026-09-09
fix(lib): don't require full crypto feature for macho-module

CfirTsabari:
## Summary

`macho-module`'s Cargo feature requested the entire `crypto` feature
bundle, even though the `macho` module never performs signature
verification — it only computes MD5/SHA-1 fingerprints
(`lib/src/modules/macho/mod.rs`) and parses embedded CMS certificates
structurally to read issuer/subject/thumbprint
(`lib/src/modules/utils/asn1.rs`), never validating a certificate
chain or verifying a signature.

This PR lists macho-module's actual dependencies explicitly instead of
requesting `crypto`, so an embedder who only builds `macho-module` no
longer pulls in the asymmetric-crypto verification stack (`rsa`, `dsa`,
`ecdsa`, `p256`, `p384`, `md2`, `sha2`, and their transitive deps — 51
fewer crates in a standalone build).

Along the way I found `utils::leb128` (a small LEB128 varint parser,
`nom`-only, no crypto at all) was gated behind `feature = "crypto"`
rather than the modules that actually use it (`macho-module` and
`dex-module`). That's a latent bug independent of the above:
`dex-module` built by itself (`cargo build -p yara-x
--no-default-features --features dex-module`) already fails today on
`main`, only masked in practice because `macho-module` happened to
also request `crypto`. Fixed by gating `leb128` on
`any(macho-module, dex-module)` instead.

## Testing

- `cargo build -p yara-x --no-default-features --features macho-module` — now succeeds (previously failed once `crypto` was dropped from the feature list)
- `cargo build -p yara-x --no-default-features --features dex-module` — now succeeds standalone (previously already broken on `main`)
- `cargo build -p yara-x --no-default-features --features pe-module` / `crx-module` — unaffected, still pull `crypto` as before
- `cargo build -p yara-x` (default features) — unaffected
- `cargo test -p yara-x --lib modules::macho modules::dex modules::pe modules::crx` — all passing
- `cargo check` across the whole workspace (capi, py, cli, ls, js-wasm) — passing
- `cargo tree -i rsa` / `-i dsa` confirm macho-module alone no longer pulls the verification crates, while `der-parser`/`x509-parser` remain (still needed for macho's own certificate parsing)

No behavior change for any existing consumer: `default-modules` still
enables `crypto` through `pe-module`/`dotnet-module`/`crx-module`, so
default builds are identical.


lib/Cargo.toml
```diff
@@ -221,9 +221,14 @@ lnk-module = [
 
 # The `macho` module parses Mach-O files.
 macho-module = [
-    "crypto",
+    "dep:const-oid",
```
plusvic, line 224: The intention with the `crypto` bundle is preventing cases like this where you have to enumerate a bunch of dependencies. The `asn1` Rust module, which is used by the `macho` module, requires many of the crates included in the `crypto` bundle. 

If this PR is accepted, any module that depends on `asn1` for parsing some data structure, would need to explicitly depend on `der-parser`, `x506-parser`, `digest`, `sha1`, etc. That's what `crypto` intends to avoid.

It's true that `crypto` also includes other crates that are not required by `macho`, but honestly I don't see a lot of value in allowing such a fine-grained control over the crates that are included when you choose one module or another. What's your use-case for this?

CfirTsabari: for the use case: an embedder who links macho-module without needing signature verification at all currently has no way to avoid ~51 crates of RSA/DSA/ECDSA machinery, so I thought it make sense to allow opting in accordingly.
I can change the pr to introduce a new feature called asn1, which will be used by crypto and macho-module, this feature point to the needed features, does that make sense?

plusvic: But is this a theoretical use-case? Or real one? I understand that someone may not be interested in signature verification, and by removing all those dependencies the final binary would be a bit smaller and compilation time a bit lower. But unless this brings a real benefit for someone I don't see the need of adding a new feature.

CfirTsabari: So no it's not a real use case, my use case is removing crypto, but when I investigated the code, I noticed this along over issues, such as #766 and #767 which I do need for my use case also #768 which i don't need for my use case. I thought it might be a "good samaritan" move  to try and address or raise the issues that doesn't affect my use case.

plusvic: Ok. For the moment I would prefer not introducing this change. It adds yet another feature, and honestly, I don't want to encourage people to build YARA-X without crypto support because this introduces a mismatch between the features supported by one build and the other. In the past, these mismatches has been the source of confusion for users of the module that was compiled without crypto support.

PR #767, opened 2026-09-10
fix(capi): add per-module feature forwards, matching yara-x-py

CfirTsabari:
`capi/Cargo.toml` forwarded only `magic-module`, so C/C++ embedders —
and Go, whose bindings go through capi — had no way to slim a build by
disabling unneeded modules; `yara-x = { workspace = true }` always
pulled in `yara-x/default`, including every default module.

Add a `<name>-module` feature per module (mirroring yara-x-py) plus a
`default-modules` group that reproduces exactly what `default-modules`
already enabled, so existing consumers see no change. Declare the
`yara-x` dependency directly (path + version) instead of inheriting it
from the workspace table, with `default-features = false` and the four
non-module defaults forwarded explicitly — same treatment as
yara-x-py's dependency, and for the same reason: overriding
`default-features` on a `workspace = true` dependency is only allowed
if the workspace's own `[workspace.dependencies]` entry already sets
it, and changing that shared entry would also affect `cli`, `js-wasm`
and `ls`.

Verified: default build of yara-x-capi is unchanged (675 crates, `rsa`
and `tinyzip` both present via default-modules); building with only
`console-module,math-module,string-module,time-module` now genuinely
excludes both (466 crates, `rsa`/`tinyzip` absent from the tree) —
previously there was no way to express this at all.


plusvic:
Already fixed in cbca186.

